PR #3, opened 2020-11-30
Add properties search

benbaumann95:
Description of change
Probably best to go through commits..
But in summary:

Adds relevant frontend libraries
Adds axios to talk to our API
Adds jest unit testing
Add necessary components to render Properties table and view pages
Adds very basic error handling
Add PropType validations
Add Router (https://nextjs.org/docs/api-reference/next/router) so we can have pages and routing


styles/all.scss
```diff
@@ -0,0 +1 @@
+@import 'node_modules/govuk-frontend/govuk/all';
```
EGiataganas: 👍

utils/api/properties.js
```diff
@@ -0,0 +1,21 @@
+import axios from 'axios'
+
+console.log(process.env.ENDPOINT_API)
```
NeilvB: Can we ✂️ this?

NeilvB (approved):
Just a `console.log` I think we can remove

utils/api/properties.js
benbaumann95: Ah might have been left in an older commit. Should be removed in the most up to date commit but thanks for flagging

NeilvB: Ah of course, yep it is. Thanks

utils/api/properties.js
```diff
@@ -0,0 +1,19 @@
...
+  return data
+}
+
+export const getProperty = async (propertyReference) => {
+  const { data } = await axios.get(
+    `${NEXT_PUBLIC_PROPERTIES_ENDPOINT_API}/${propertyReference}`
+  )
+
+  return data
+}
```
EGiataganas, line 19: Is there any particular reason for deleting the `/properties` and required it as part of the environment variable? Isn't better to have only one env var which will be the `API_URL`?

EGiataganas (approved):

PR #7, opened 2020-12-07
Ip whitelist

benbaumann95:
### Description of change

For now, temporarily add whitelisting of IP addresses before we implement auth solution

### DIFF
--- a/src/pages/_app.js
+++ b/src/pages/_app.js
@@ -2,6 +2,8 @@ import '../styles/all.scss'
 import App from 'next/app'
 import Layout from '../components/Layout'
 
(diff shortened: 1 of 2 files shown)


src/pages/_app.js
```diff
@@ -2,6 +2,10 @@ import '../styles/all.scss'
 import App from 'next/app'
 import Layout from '../components/Layout'
 
+const { NEXT_PUBLIC_ENDPOINT_API } = process.env
```
EGiataganas, line 5: Where is this used?

EGiataganas (approved):

PR #9, opened 2020-12-07
Add cypress e2e/integration testing library

benbaumann95:
### Description of change

Adds e2e/integration testing

### DIFF
--- a/cypress/integration/search_for_property.spec.js
+++ b/cypress/integration/search_for_property.spec.js
@@ -1,0 +1,70 @@
+/// <reference types="cypress" />
+
+beforeEach(() => {
+  cy.fixture('properties/properties.json').as('propertiesList')
+  cy.server()
+})
+
+describe('Search for property', () => {
+  it('Search for property by postcode', () => {
+    cy.fixture('properties/property.json').as('property')
+
+    // Stub request for search on properties by postcode
+    cy.route('GET', 'api/v2/properties/?q=e9 6pt', '@propertiesList')
+
+    // Search by postcode
+    cy.visit('/')
+    cy.get('.govuk-input').clear().type('e9 6pt')
+    cy.get('[type="submit"]').contains('Search').click()
+    cy.url().should('contains', 'properties/search?q')
+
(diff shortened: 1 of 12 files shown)


cypress/integration/search_for_property.spec.js
```diff
@@ -0,0 +1,60 @@
...
+
+beforeEach(() => {
+  // Use properties fixture as mock stubbed data
+  cy.fixture('properties/properties.json').as('propertiesList')
+  cy.fixture('properties/property.json').as('property')
+  cy.server()
+})
+
+describe('Search for property', () => {
+  it('Search for property by postcode', () => {
```
NeilvB, line 11: Should we split out part of this into a spec for searching by address?

benbaumann95: Yeah although maybe just the initial search as there will be duplication from above when going into the actual property. Unless that is OK?

NeilvB: Yes I was thinking along same lines as you, and minimal duplication - just check the address search is possible and has the check for number of results, but no more steps than that.

cypress/integration/search_for_property.spec.js
```diff
@@ -0,0 +1,60 @@
...
+    cy.visit('http://localhost:5000/', { timeout: 3000 })
+
+    // Stub request for search on properties by address
+    cy.route('GET', 'api/v2/properties/?q=pitcairn', '@propertiesList')
```
NeilvB, line 15: My IDE prodded me to say that `.route()` is deprecated in favour of [.intercept()](https://docs.cypress.io/api/commands/intercept.html#Comparison-to-cy-route) - do you think that would work for us? It looks like it might be a small change.

NeilvB (approved):

cypress/integration/search_for_property.spec.js
benbaumann95: Hey yes I'm going to address this in a next PR thanks for flagging!

PR #14, opened 2020-12-11
Tweak frontend to match backend

EGiataganas:
### Description of change

- Fix eslintrc with `prettier:fix`
- Rename CautionaryAlerts to LocationAlert
  - We've updated the API to use HACT standard
  - We've changed the cautionary alerts array to be an alert hash with locationAlert and personAlert array.
  - We've renamed alertCode to code and description to comments
- Add person alerts
- Wrap location and person alerts to the same class
- Add tenure description

### Screenshot
Screenshot 2020-12-14 at 17 23 40

There's a tiny style difference where cautionary alerts and tenure should be a list which will happen on a later PR


src/components/Property/PersonAlerts.js
```diff
@@ -0,0 +1,30 @@
...
+  let alertsToShow = (prsnAlerts) => {
+    if (prsnAlerts.length != 0) {
+      let alertsHtml = prsnAlerts.map((alert, index) => {
+        return (
+          <li key={index}>
+            {alert.comments} (
+            <span className="govuk-!-font-weight-bold">{alert.code}</span>)
+          </li>
+        )
+      })
```
siapankina, line 13: 👍

siapankina (approved):
Looks good to me ;)

cypress/fixtures/properties/property.json
```diff
@@ -13,12 +13,26 @@
...
   },
-  "cautionaryAlerts": [
-    {
-      "alertCode": "DIS",
-      "description": "Property Under Disrepair",
-      "startDate": "2011-02-16",
-      "endDate": null
-    }
-  ]
+  "alert": {
```
benbaumann95, line 16: Just FYI planning to change this on backend as you suggested to `alerts`

src/components/Property/PropertyDetails.test.js
```diff
@@ -16,17 +16,28 @@ describe('PropertyDetails component', () => {
...
-          {
-            alertCode: 'DIS',
-            description: 'Property Under Disrepair',
-            startDate: '2011-02-16',
-            endDate: null,
-          },
-        ],
-      },
+    },
+    alert: {
```
benbaumann95, line 20: same as above

benbaumann95 (approved):
Looks great. Could we also get a screenshot perhaps of what it looks like? Would be good to get one of the designers to have a look and approve the UI

src/components/Property/Tenure.js
```diff
@@ -0,0 +1,22 @@
+import PropTypes from 'prop-types'
+
+const Tenure = ({ tenure }) => {
+  let tenureToShow = (tenure) => {
+    if (tenure) {
+      return (
+        <ul className="govuk-tag bg-turquoise">
```
benbaumann95, line 7: Hey where is this `bg-turquoise` being defined in css? I think you may have missed it out 😄

PR #17, opened 2020-12-16
Make styling changes to match designs

benbaumann95:
### Description of change

Update UI of Tenure and Alerts to match designs (dependent on https://github.com/LBHackney-IT/repairs-api/pull/23)

- Change `code` to `type` on alerts with previous API change
- Rearrange order of tenure/alerts inline with Repairs Hub V1 (using this as reference https://github.com/LBHackney-IT/repairs-management/blob/develop/app/views/shared/_tenure_and_cautionary_contact_info.html.erb)
- Update integration specs to match updated UI
- Add case for when tenure is not raisable and render correct class
- Update fixtures with canRaiseRepair
- Update property object from property.test with updated response object with alerts and tenures

Fix bug for when tenure returns null from the API to still render components

- Set tenure to empty object {} when null is returned
- Safely check if there is a typeCode on tenure before rendering component
- Add integration spec for this scenario

### **Previous version:**
Screenshot 2020-12-16 at 12 38 05

### **Updated version:**
Screenshot 2020-12-16 at 12 33 25

Updated version now matches screenshot in trello card: https://trello.com/c/No3ca2yv (Screenshot 2020-10-08 at 15.03.29 copy.jpg)

### DIFF
--- a/cypress/integration/search_for_property.spec.js
+++ b/cypress/integration/search_for_property.spec.js
@@ -3,6 +3,7 @@
 beforeEach(() => {
   cy.fixture('properties/properties.json').as('propertiesList')
   cy.server()
+  cy.visit('/')
 })
 
 describe('Search for property', () => {
@@ -13,10 +14,9 @@ describe('Search for property', () => {
     cy.route('GET', 'api/v2/properties/?q=e9 6pt', '@propertiesList')
 
     // Search by postcode
-    cy.visit('/')
     cy.get('.govuk-input').clear().type('e9 6pt')
     cy.get('[type="submit"]').contains('Search').click()
-    cy.url().should('contains', 'properties/search?q')
+    cy.url().should('contains', 'properties/search?q=e9%25206pt')
 
     cy.get('.govuk-heading-s').contains(
       'We found 2 matching results for: e9 6pt'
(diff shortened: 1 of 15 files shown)


cypress/integration/search_for_property.spec.js
```diff
@@ -71,12 +67,37 @@ describe('Search for property', () => {
...
-    cy.get('.lbh-header__title-link').click()
     cy.get('.govuk-input').type('pitcairn')
     cy.get('[type="submit"]').contains('Search').click()
 
     cy.get('.govuk-heading-s').contains(
       'We found 2 matching results for: pitcairn'
     )
   })
+
+  it('Search for property with unraisable repair', () => {
```
NeilvB, line 78: Maybe '... with unraisable tenure code'?

benbaumann95: I had a think about this and was wondering what  does the unraisable refer to if it's not with a repair context. But I suppose the app name makes it implicit 😁

NeilvB: I wonder if this is telling us we'd benefit from splitting out a new integration spec. 

The last couple of specs here don't perform searches. What do you think about having a `show_property.spec.js` or similar?  Then we could have a `show property with a tenure type that permits repair raising` and its opposite. We might also want to bring some of the more detailed checks from 'Search for property' into its own spec there too. I'm thinking it would be a bit more readable / concise to have our search spec just input search terms, do some checks on the table results and then a check that the link in the table takes us to the show page, but no more. The rest is left to our show page spec. What do you think?

siapankina (approved):
Well done ⭐

cypress/integration/search_for_property.spec.js
benbaumann95: Hey thanks for the suggestion @NeilvB . It makes a lot of sense, I'm definitely in agreement with you. Will make those changes 👍

NeilvB (approved):

EGiataganas (approved):

PR #18, opened 2020-12-16
Add raise a repair line 

siapankina:
### Description of change

Add raise repair line on property show page to check if we are liable for repairs at this property
Implement css 
Update code to be sure that the tenure's propType is always an object 

### Story Link

https://trello.com/c/3yBCDiDd/157-as-a-user-i-need-to-know-if-we-are-liable-for-repairs-at-this-property-so-i-dont-raise-a-repair-unnecessarily

ScreenShots: 

Screenshot 2020-12-17 at 13 06 58
Screenshot 2020-12-17 at 13 07 07


src/components/Property/PropertyDetails.js
```diff
@@ -2,52 +2,81 @@ import PropTypes from 'prop-types'
...
+      )
+    }
+  }
+  return (
+    <div>
+      <h1 className="govuk-heading-l">
+        {hierarchyType.subTypeDescription}: {address.addressLine}
+      </h1>
+      <div>
+        {repairsStatus(tenure.canRaiseRepair, hierarchyType.subTypeDescription)}
```
NeilvB, line 45: I like our `canRaiseRepair`  property name and I feel like it's almost asking us to be used in an `if` condition directly for a bit of a readability boost, rather than passed down into our helper-type function here.

With a couple of smaller functions, I wonder if this approach might be a bit more readable:

```
{
  if(tenure.canRaiseRepair) {
    renderRaiseRepairLink(hiearchyType.subTypeDescription)
  } else {
    renderRepairUnavailableText()
  }
}
```

NeilvB (commented):
Looks great. I just added a small idea for a function refactor.

src/components/Property/PropertyDetails.js
benbaumann95: I'd agree with Neil and also perhaps extract this function into its own component. For readability, it helps that the file doesn't get too big and then you can pass down the state of whatever you need as props

src/components/Property/PropertyDetails.js
```diff
@@ -2,52 +2,81 @@ import PropTypes from 'prop-types'
...
-            </span>
+}) => {
+  let repairsStatus = (status, subTypeDescription) => {
+    if (status) {
+      return (
+        <span class="govuk-heading-m">
+          <strong>
+            <Link href="">
+              <a>Raise a repair on this {subTypeDescription.toLowerCase()} </a>
+            </Link>
```
benbaumann95, line 21: Did we say to have a link or just text for now? It might be worth checking with Sarah but it could be a bit misleading to have a link that isn't functional

src/styles/components/header.scss
```diff
@@ -1,3 +1,7 @@
 .lbh-header__service-name {
   font-family: $govuk-font-family;
 }
+
+.govuk-heading-l {
+  margin-bottom: 10px !important;
+}
```
benbaumann95, line 7: I would add another class for doing this. This would mean anytime we use `.govuk-heading-l` it would have this overriding margin and we should try to avoid `!important` if possible

src/components/Property/Tenure.js
```diff
@@ -2,7 +2,7 @@ import PropTypes from 'prop-types'
 
 const Tenure = ({ tenure }) => {
   let tenureToShow = (tenure) => {
-    if (tenure?.typeCode) {
+    if (tenure.typeDescription) {
```
benbaumann95, line 5: Will this blow up if `tenure` is `{}` ? Safe navigator  in ruby `&.` has the js equivalent of `?.`

siapankina: I checked with` {}` it works fine 
I checked in console, when you do `tenure.typeDescription` it returns `false` if we have empty object 
I guess that's the way to check it in js

benbaumann95: Interesting... I just opened my console and tried `{}.bla` and you get an error. I wonder what's going on here that's different

benbaumann95: Ah you have to declare the object first. So it would return `undefined` and that interprets it as false

src/components/Property/PropertyDetails.js
```diff
@@ -2,52 +2,64 @@ import PropTypes from 'prop-types'
...
+      return <RepairUnavailableText />
+    }
+  }
+  return (
+    <div>
+      <h1 className="govuk-heading-l repair-laber">
+        {hierarchyType.subTypeDescription}: {address.addressLine}
+      </h1>
+      <div>
+        {labelToShow(tenure.canRaiseRepair, hierarchyType.subTypeDescription)}
```
NeilvB, line 28: The components you've added look great.

This is nitpicky of me but please could we add lines 16-20 in here directly? It's a small change but I think in this case it should add a bit more readability. Breaking things down into small functions works really nicely in lots of places but here we're not getting much benefit from that function IMO. It's hard to accurately name what the function does which I think is nudging us in a direction - I know what you mean about 'label' but really it's a kind of 'linkOrText' function, and if we're at that stage I feel like just having the conditional in here straight up is better.

cypress/integration/show_property.spec.js
```diff
@@ -5,12 +5,17 @@ beforeEach(() => {
 })
 
 describe('Show property', () => {
-  it('Display property details, tenure and alerts', () => {
+  it('Displays property details, tenure, alerts and permit to raise a repair', () => {
```
NeilvB, line 8: 👍

src/components/Property/PropertyDetails.js
NeilvB: As you explained in Slack, it sounds like I'm misunderstanding parts of how React works here with conditional rendering. I'll resolve and maybe we can revisit another time once we learn more - if applicable. Thank you for pointing that out.

src/components/Property/PropertyDetails.js
```diff
@@ -2,52 +2,61 @@ import PropTypes from 'prop-types'
...
+    <div>
+      <h1 className="govuk-heading-l repair-laber">
+        {hierarchyType.subTypeDescription}: {address.addressLine}
+      </h1>
+      <div>
+        {tenure.canRaiseRepair ? (
+          <RaiseRepairLink description={hierarchyType.subTypeDescription} />
+        ) : (
+          <RepairUnavailableText />
+        )}
```
benbaumann95, line 25: I'd have this like `<RaiseRepairStatus>` and have all this logic in that component. No need for two components here I don't think. Especially as the `<RepairUnavailableText />` holds no state

src/components/Property/RepairUnavailableText.js
```diff
@@ -0,0 +1,13 @@
+const RepairUnavailableText = () => (
+  <div class="govuk-warning-text">
+    <span class="govuk-warning-text__icon" aria-hidden="true">
+      !
+    </span>
+    <strong class="govuk-warning-text__text">
+      <span class="govuk-warning-text__assistive">Warning</span>
```
EGiataganas, line 7: this should be `className`

src/components/Property/RaiseRepairLink.js
```diff
@@ -0,0 +1,13 @@
+import PropTypes from 'prop-types'
+
+const RaiseRepairLink = ({ description }) => (
+  <span class="govuk-heading-m text-green">
+    <strong>Raise a repair on this {description.toLowerCase()}</strong>
+  </span>
+)
+
+RaiseRepairLink.propTypes = {
+  tenure: PropTypes.string.isRequired,
```
EGiataganas, line 10: this should be description

src/components/Property/__snapshots__/PropertyDetails.test.js.snap
```diff
@@ -4,10 +4,19 @@ exports[`PropertyDetails component should render properly 1`] = `
 <DocumentFragment>
   <div>
     <h1
-      class="govuk-heading-l"
+      class="govuk-heading-l repair-laber"
```
benbaumann95, line 7: typo?

src/styles/components/header.scss
```diff
@@ -1,3 +1,7 @@
 .lbh-header__service-name {
   font-family: $govuk-font-family;
 }
+
+.repair-laber {
```
benbaumann95, line 5: same as above

cypress/integration/show_property.spec.js
```diff
@@ -5,12 +5,17 @@ beforeEach(() => {
...
     // Stub request with property response
     cy.fixture('properties/property.json').as('property')
     cy.route('GET', 'api/v2/properties/00012345', '@property')
     cy.visit('properties/00012345')
 
+    // can raising a repare
```
benbaumann95, line 14: Type here should be `repair` instead of `repare`

benbaumann95: And maybe `can raise a repair`

benbaumann95 (approved):
🎉

PR #22, opened 2020-12-18
Add cypress-audit for lighthouse automation on e2e tests

benbaumann95:
### Description of change

- Generate reports folder on the fly and add lighthouse report files for each url as JSON
- Add cypress/reports into circleCI artefacts to retrieve the JSON report
- Paste JSON into https://googlechrome.github.io/lighthouse/viewer/ for full report with UI and metrics breakdown
- Set accessibility score at 90
- Use headless chrome to run rests to support lighthouse audit
- Run yarn e2e:audit to run with accessibility testing

### Story Link

https://trello.com/c/SKIacteL/274-add-automated-accessibility-checks-to-our-build-process

### **Folders and JSON files generated on the fly as you run the e2e tests:**
Screenshot 2021-01-04 at 17 40 46

### **Lighthouse Error in cypress interactive test**
Screenshot 2020-12-18 at 13 15 32

### **CircleCI error if tests don't meet threshold as defined in `cypress.json`**
Screenshot 2020-12-18 at 13 32 42

### **Reports saved to artifcacts**
Screenshot 2021-01-04 at 18 09 36

### **Lighthouse Report Viewer UI https://googlechrome.github.io/lighthouse/viewer/**
Screenshot 2020-12-18 at 13 22 44

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -1,4 +1,11 @@
 {
+  "env": {
+    "runAudit": true
+  },
   "defaultCommandTimeout": 10000,
-  "baseUrl": "http://localhost:5001/"
+  "baseUrl": "http://localhost:5001/",
+  "video": false,
+  "lighthouse": {
+    "accessibility": 90
+  }
 }
--- a/cypress.json
+++ b/cypress.json
@@ -1,7 +1,4 @@
 {
-  "env": {
-    "runAudit": true
-  },
   "defaultCommandTimeout": 10000,
   "baseUrl": "http://localhost:5001/",
   "video": false,
--- a/package.json
+++ b/package.json
@@ -13,8 +13,10 @@
     "test": "jest",
     "lint": "eslint . --max-warnings 0 --ext .js,.jsx && echo 'Lint complete.'",
     "e2e": "PORT=5001 server-test start 5001 cy:run",
-    "interactive:e2e": "PORT=5001 server-test start 5001 cy:open",
+    "e2e:audit": "PORT=5001 server-test start 5001 cy:run:audit",
+    "e2e:interactive": "PORT=5001 server-test start 5001 cy:open",
     "cy:run": "cypress run --browser chrome --headless",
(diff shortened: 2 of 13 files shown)


cypress.json
```diff
@@ -1,4 +1,11 @@
 {
+  "env": {
+    "runAudit": true
+  },
   "defaultCommandTimeout": 10000,
-  "baseUrl": "http://localhost:5001/"
+  "baseUrl": "http://localhost:5001/",
+  "video": false,
+  "lighthouse": {
+    "accessibility": 90
```
siapankina, line 9: I think we were going with `">75"`
maybe will be better to check this with Sarah

siapankina (approved):
Great job!

cypress.json
NeilvB: I suggested we use that threshold of 75. But that was just as an initial aim and open for revision. I'm happy for us to go with 90 as we can reassess later.

benbaumann95: Hey, I had a good look around other projects that use some sort of accessibility reporting and the benchmark they use is generally 95+ so I thought we should be aiming for at least 90. It will be good practice to add any accessibility tweaks from an early stage and it seems so far that we are way above this 75 mark anyway

package.json
```diff
@@ -13,8 +13,10 @@
     "test": "jest",
     "lint": "eslint . --max-warnings 0 --ext .js,.jsx && echo 'Lint complete.'",
     "e2e": "PORT=5001 server-test start 5001 cy:run",
-    "interactive:e2e": "PORT=5001 server-test start 5001 cy:open",
-    "cy:run": "cypress run",
+    "e2e:audit": "PORT=5001 server-test start 5001 cy:run:audit",
+    "e2e:interactive": "PORT=5001 server-test start 5001 cy:open",
```
benbaumann95, line 17: Better script naming like this imo

NeilvB (approved):
Really great to have this added to the project

PR #23, opened 2021-01-05
Cache API requests

siapankina:
-updated tests
-set maxTime to 5 min
-moved GET requests logic for properties into ApiRequestsCacher folder,
	where API responses are being cached
-added dependancy ` lru cache`  https://github.com/isaacs/node-lru-cache

***Moving API/REPAIRS under Pages folder
I would suggest doing that after this PR is approved, then create a new PR just for that

### Story Link

https://trello.com/c/OtOaPFgl/284-cache-api-requests-on-frontend-nextjs-app


siapankina:
Closing this pull request because we don'y need it right now , might be useful in future

PR #26, opened 2021-01-12
Route all service API requests through NextJS

NeilvB:
### Description of change

Communicate with the service API via intermediary Next JS [API routes](https://nextjs.org/docs/api-routes/introduction).

### Notable changes

- [Move URL and token vars to the server side](.env.sample)
- Adds a frontend-api-client directory under utils with helpers to
  call the Next JS API
- Adds a service-api-client directory under utils with helpers,
  similar to the above, to call the service API

### Story Link

https://trello.com/c/kl5SPQtP/292-rework-clientside-service-api-calls-to-hackney-convention-using-api-routes

### Known issues

The new Next JS API endpoints contain a lot of duplicate logic:

- [/api/properties.js](src/pages/api/properties.js)
- [api/properties/[id].js](src/pages/api/properties/[id].js)

An improvement to our API route structure would be to have a "catch-all" endpoint which takes all requests, breaks apart their paths and params and uses those in the service API call dynamically. I feel that adding this overall structure to the app is more important at the moment, so I suggest this can be added later.


cypress/integration/search_for_property.spec.js
```diff
@@ -11,7 +11,7 @@ beforeEach(() => {
 describe('Search for property', () => {
   it('Search for property by postcode', () => {
     // Stub request for search on properties by postcode
-    cy.route('GET', 'api/v2/properties/?q=e9 6pt', '@propertiesList')
```
benbaumann95, line 14: Where did the v2 go ? 😀

src/components/Property/PropertyView.js
```diff
@@ -3,7 +3,7 @@ import { useState, useEffect } from 'react'
 import PropertyDetails from './PropertyDetails'
 import Spinner from '../Spinner/Spinner'
 import ErrorMessage from '../Errors/ErrorMessage/ErrorMessage'
-import { getProperty } from '../../utils/api/repairs/properties/properties'
+import { getProperty } from '../../utils/frontend-api-client/repairs'
```
benbaumann95, line 6: Is there a reason to have this function under `/repairs` rather than `/properties` ?
For example, when we post data from the repairs form or get a work order would we expect this to sit in a different file?

benbaumann95: Or maybe under `.../repairs/properties` would be a good option?

src/pages/api/properties.js
```diff
@@ -0,0 +1,12 @@
+import { getProperties } from '../../utils/service-api-client/repairs'
+
+export default async function endpoint(req, res) {
```
benbaumann95, line 3: Just fyi I think latest syntax would be `export default async (req, res) => {}` but not a big deal 😄

src/pages/api/properties/[id].js
```diff
@@ -0,0 +1,12 @@
+import { getProperty } from '../../../utils/service-api-client/repairs'
+
+export default async function endpoint(req, res) {
```
Comment (benbaumann95) [inline src/pages/api/properties/[id].js:3]: same comment as above

benbaumann95 (commented):
Hey this looks awesome and should allow us to now connect with the api securely 🎉 
Just left a few minor comments mostly around naming which i'm happy to discuss

src/components/Property/PropertyView.js
NeilvB: Makes much more sense! [Moved under `/properties`](https://github.com/LBHackney-IT/repairs-hub-frontend/pull/26/commits/35b2262c25a2b8c62917330d05a31eeec5f9fcfa#diff-db659b9529a80962294d5249575a862de8eac3ecee843705b9720b9562182ce5)

src/pages/api/properties.js
NeilvB: Thanks. [Updated](https://github.com/LBHackney-IT/repairs-hub-frontend/pull/26/commits/35b2262c25a2b8c62917330d05a31eeec5f9fcfa#diff-f60e89599c96f95b247f9e10b8008733afccbe32faff1ee0ad7ffb502c67ae39R3)

src/pages/api/properties/[id].js
Comment (NeilvB) [reply src/pages/api/properties/[id].js]: Thanks. [Updated](https://github.com/LBHackney-IT/repairs-hub-frontend/pull/26/commits/35b2262c25a2b8c62917330d05a31eeec5f9fcfa#diff-878483077f770b4fd8cf58fcd922796141740d9db395bbb0424d24b7ba955a34R3)

cypress/integration/search_for_property.spec.js
NeilvB: This is now mocking out the response from the Next JS API rather than the service API. But this has made me realise I'm missing a layer of tests - for the API routes (not just the util functions they call). I'll add those.

benbaumann95 (approved):
Adding those tests sounds good! Assume you also checked that no token is exposed in the browser when you were testing on staging last night? Otherwise good to go 🚀

src/pages/api/properties.js
```diff
@@ -0,0 +1,12 @@
...
+export default async (req, res) => {
+  try {
+    const { status, data } = await getProperties(req.query)
+
+    res.status(status).json(data)
+  } catch (err) {
+    console.log(err)
+    return { status: 500 }
+  }
+}
```
EGiataganas, line 12: Do you think this should be `pages/api/properties/index.js`?

cypress/integration/search_for_property.spec.js
NeilvB: Will add in a separate PR as discussed.

src/pages/api/properties.js
NeilvB: I see what you mean. We could do. The docs don't seem to recommend one way over another. https://nextjs.org/docs/api-routes/dynamic-api-routes#index-routes-and-dynamic-api-routes

PR #27, opened 2021-01-13
[WIP]Contractor job page

siapankina:
### Description of change

 Contractors can see all job orders
 created dummy data to be able to display jobs with descriptions 
 added tests to check that correct data is shown  

### Story Link

https://trello.com/c/jBQjau0m/288-as-an-contractor-admin-i-need-to-see-jobs-assigned-to-me-so-that-i-can-assign-them-to-operatives

Screenshot 2021-01-13 at 14 58 59


cypress/fixtures/work_orders/jobs.json
```diff
@@ -0,0 +1,29 @@
...
+	},
+	{
+		"reference": "00012345",
+		"dateRaised": "2021-01-03T16:24:26.632",
+		"lastUpdated": "2021-01-12T16:24:26.632",
+		"priority": "N - Normal",
+		"propery": "1-53 Pitcairn House St Thomass Square",
+		"owner": "",
+		"description": ""
+	}
```
siapankina, line 28: the indentation looks good on my machine 🤔

cypress/integration/show_work_orders.spec.js
```diff
@@ -0,0 +1,62 @@
...
+      cy.contains('th', 'Reference')
+      cy.contains('th', 'Date raised')
+      cy.contains('th', 'Last update')
+      cy.contains('th', 'Priority')
+      cy.contains('th', 'Property')
+      cy.contains('th', 'Description')
+    })
+
+    //can see reference number
+    cy.get('.govuk-table__cell').within(() => {
```
benbaumann95, line 28: Instead of repeating `cy.get('.govuk-table__cell')` for all these examples could you just have all the `cy.contains(...)` in this one block instead? e.g. 
```
cy.get('.govuk-table__cell').within(() => {
  cy.contains(...)
  cy.contains(...)
}
```

cypress/integration/show_work_orders.spec.js
```diff
@@ -0,0 +1,62 @@
...
+    })
+
+    //can see date raised
+    cy.get('.govuk-table__cell').within(() => {
+      cy.contains('13 January 2021')
+    })
+
+    //can see last updated
+    cy.get('.govuk-table__cell').within(() => {
+      cy.contains('13 January 2021')
```
NeilvB, line 39: I think this should be different to the date raised (above). 

Ideally we'd scope all of our searches to within a particular column of the table. I'd need to look up how to do that in Cypress.

src/components/WorkOrders/JobRow.js
```diff
@@ -0,0 +1,41 @@
...
+  dateRaised,
+  lastUpdated,
+  priority,
+  propery,
+  description,
+}) => (
+  <tr className="govuk-table__row govuk-table__row--clickable govuk-body-s">
+    <td className="govuk-table__cell">{reference}</td>
+    <td className="govuk-table__cell">
+      {dateRaised ? dateToStr(dateRaised) : '—'}
```
NeilvB, line 21: It would be good to have a component test for this, especially so we can test logic this like.

src/utils/service-api-client/work-orders.js
```diff
@@ -0,0 +1,15 @@
+import axios from 'axios'
+
+const { DOTNET_SERVICE_API_URL } = process.env
+const { DOTNET_SERVICE_API_KEY } = process.env
```
benbaumann95, line 4: You can rebase with main now that we have updated the api and key

src/utils/frontend-api-client/work-orders.js
```diff
@@ -0,0 +1,7 @@
+import axios from 'axios'
+
+export const getWorkOrders = async () => {
+  const { data } = await axios.get('/api/work-orders')
```
benbaumann95, line 4: I think the idea is to match these endpoints with the service api endpoint. So if you look on swagger i think this exists now as `/repairs`

src/styles/colours.scss
```diff
@@ -16,3 +17,8 @@ $repairs-hub-colours: (
 .text-green {
   color: repairs-hub-colour('housing-2');
 }
+
+.work-order-hours {
```
benbaumann95, line 21: probably doesn't belong in this file if it's not an actual colour definition

src/pages/api/work-orders.js
```diff
@@ -0,0 +1,12 @@
...
+export default async (req, res) => {
+  try {
+    const { status, data } = await getWorkOrders()
+
+    res.status(status).json(data)
+  } catch (err) {
+    console.log(err)
+    return { status: 500 }
+  }
+}
```
benbaumann95, line 12: You will need to add auth stuff  like with the other files in api/pages

src/components/WorkOrders/JobRow.js
```diff
@@ -0,0 +1,50 @@
+import PropTypes from 'prop-types'
+
+const dateToStr = (date) =>
+  date.toLocaleString('en-GB', {
+    year: 'numeric',
+    month: 'long',
+    day: 'numeric',
+  })
+
+const hourTosTr = (date) =>
```
benbaumann95, line 10: This const name is a bit confusing as it suggests converting hours to a string but i think you are converting a date right?

benbaumann95 (commented):
Do we need to add another card to add the  scoping to contractors? I assume we've changed the AC of this PR to just have the work orders visible globally?

src/utils/frontend-api-client/work-orders.js
NeilvB: Yes having the paths on our side match with the service API will make it easier to have a catch-all API route on our Node server. (It just "forwards" the path, params, method, body etc. on to the service API, and doesn't need to maintain any path mappings or things like that.)

src/utils/frontend-api-client/work-orders.js
```diff
@@ -0,0 +1,7 @@
+import axios from 'axios'
```
NeilvB, line 1: I think this file has been superseded by src/utils/frontend-api-client/repairs.js now and can be deleted.

src/utils/service-api-client/work-orders.js
```diff
@@ -0,0 +1,14 @@
+import axios from 'axios'
+
+const { REPAIRS_SERVICE_API_URL, REPAIRS_SERVICE_API_KEY } = process.env
+
+export const getRepairs = async () => {
```
NeilvB, line 5: Please could we move this into the repairs.js file in the same directory

src/styles/all.scss
```diff
@@ -9,3 +9,4 @@ $govuk-font-family: 'GDS Transport', arial, sans-serif !important;
 @import 'components/header';
 @import 'components/property-details';
 @import 'components/form';
+@import 'components/work-orders.scss';
```
benbaumann95, line 12: Don't need the `.scss` here

siapankina: I do need it, because in file `"work-orders.scss"` I am styling hours

cypress/integration/show_work_orders.spec.js
```diff
@@ -0,0 +1,49 @@
...
+import 'cypress-audit/commands'
+
+beforeEach(() => {
+  cy.getCookies().should('be.empty')
+  cy.setCookie('hackneyToken', Cypress.env('GSSO_TEST_KEY'))
+  cy.getCookie('hackneyToken').should(
+    'have.property',
+    'value',
+    Cypress.env('GSSO_TEST_KEY')
+  )
```
EGiataganas, line 12: can you please use the login function? as well as move this `beforeEach` block inside the `describe` block

src/components/WorkOrders/__snapshots__/JobRow.test.js.snap
```diff
@@ -0,0 +1,13 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`JobRow component should render properly 1`] = `
+<DocumentFragment>
+  000123452021-01-12T16:24:26.632
```
NeilvB, line 5: I think we need to update the component test for this snapshot to pass in Date objects rather than strings for `lastUpdated` and `dateRaised`. At the moment this part of the snapshot just gives us the string we passed in for those values rather than our human-friendly formatted dates / times.

src/components/WorkOrders/JobsTable.js
```diff
@@ -0,0 +1,92 @@
...
+                  Description
+                </th>
+              </tr>
+            </thead>
+            {workOrders && (
+              <tbody className="govuk-table__body">
+                {workOrders.map((job, index) => (
+                  <JobRow key={index} {...job} />
+                ))}
+              </tbody>
```
EGiataganas, line 82: how about renaming this component into `JobView` and extract these lines into another component called `JobsTable`?

src/components/WorkOrders/JobView.js
```diff
@@ -0,0 +1,55 @@
...
+
+    try {
+      const data = await getRepairs()
+
+      data.forEach((job) => {
+        job.dateRaised = convertDate(job.dateRaised)
+        job.lastUpdated = convertDate(job.lastUpdated)
+      })
+
+      const sortedByDate = data.sort((a, b) => b.dateRaised - a.dateRaised)
```
EGiataganas, line 24: would be great to move this functionality of sorting data by date into `utils/date` and a test for this

cypress/fixtures/work_orders/jobs.json
```diff
@@ -0,0 +1,29 @@
+[
+  {
+    "reference": "00012345",
```
benbaumann95, line 3: All your references in this spec are the same btw, I think there is a unique constraint on them so should be different

cypress/integration/show_work_orders.spec.js
```diff
@@ -0,0 +1,38 @@
...
+      cy.contains('th', 'Reference')
+      cy.contains('th', 'Date raised')
+      cy.contains('th', 'Last update')
+      cy.contains('th', 'Priority')
+      cy.contains('th', 'Property')
+      cy.contains('th', 'Description')
+    })
+
+    //can see reference number/ date raised, last updated, priority, property address, description
+    cy.get('.govuk-table__cell').within(() => {
```
benbaumann95, line 24: Now that i'm looking at the HTML generated, shouldn't this be under `cy.get('.govuk-table__row')` ?
There are a bunch of `govuk-table__cell` classes so i'm not sure this would be the right scope to look within.
Might even be worth having the word order reference id as a data attribute so it's easier to select the relevant row with this unique identifier

cypress/integration/show_work_orders.spec.js
```diff
@@ -0,0 +1,38 @@
...
+      cy.contains('00012345')
+      cy.contains('13 Jan 2021 4:24 pm')
+      cy.contains('14 Jan 2021 10:24 am')
```
benbaumann95, line 27: In the html it looks like the time is on another line so i'm not sure the spec would pick this up as one string

src/components/WorkOrders/JobView.js
```diff
@@ -21,8 +21,7 @@ const JobView = () => {
         job.lastUpdated = convertDate(job.lastUpdated)
       })
```
EGiataganas, line 22: Do you think this could be moved inside `sortedByDate` function?

benbaumann95 (approved):

NeilvB (approved):

EGiataganas (approved):

PR #34, opened 2021-01-19
Add quantity to the raise repair form

pixeltrix:
### Description of change

Add the quantity field to the raise repair form using a free text input and validating that a whole number is entered.

### Story Link

https://trello.com/c/63HWaUZG/278-as-a-user-i-need-to-be-able-to-enter-a-quantity-relating-to-a-repair-so-that-i-can-order-the-correct-amount-of-works

### Decisions

Added some additional validations to ensure that the quantity falls in the range 1..50 - need to confirm whether this is an acceptable range. This range validation helps prevent user error entering widely incorrect values such as 1 million.

### Known issues

The error messages cause the layout to jump around - it may be better to have quantity on a separate line.

#### Valid input
image

#### Missing quantity
image

#### Invalid number
image

#### Non-whole number
image

#### Number too small
image

#### Number too large
image

### DIFF
--- a/src/components/Form/NumberInput/NumberInput.test.js
+++ b/src/components/Form/NumberInput/NumberInput.test.js
@@ -1,0 +1,33 @@
+import NumberInput from './NumberInput'
+import { fireEvent, render } from '@testing-library/react'
+
+describe('NumberInput', () => {
+  it('renders a number input', () => {
+    const inputName = 'my-number-input'
+    const inputLabel = 'My Input'
+    const { getByLabelText } = render(
+      <NumberInput name={inputName} label={inputLabel} />
+    )
+
+    const labelRegex = new RegExp(`s*${inputLabel}s*`)
+    const input = getByLabelText(labelRegex)
+
+    expect(input).toBeInTheDocument()
+    expect(input.id).toEqual(inputName)
+    expect(input.name).toEqual(inputName)
+  })
+
+  it('performs an action onChange', () => {
+    let newValue = ''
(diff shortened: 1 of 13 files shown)


src/components/Form/NumberInput/NumberInput.test.js
```diff
@@ -0,0 +1,33 @@
+import NumberInput from './NumberInput'
```
benbaumann95, line 1: Ah nice I forgot to do this. I will add some unit tests for the other Form components too 👍

src/components/Property/RaiseRepair/RaiseRepairForm.js
```diff
@@ -4,7 +4,9 @@ import { useForm } from 'react-hook-form'
 import TenureAlertDetails from '../TenureAlertDetails'
 import BackButton from '../../Layout/BackButton/BackButton'
 import { Select, Button, TextArea } from '../../Form'
+import TextInput from '../../Form/TextInput/TextInput'
```
benbaumann95, line 7: I think you could change line 6 to `import { Select, Button, TextArea, TextInput } from '../../Form'` and remove this  if you add an import and export statement to https://github.com/LBHackney-IT/repairs-hub-frontend/blob/main/src/components/Form/index.js

benbaumann95 (approved):
I think we need to override the `.govuk-input:focus` from the form text input selector so that the outline is more green in line with the colour theme of the app. I think other examples live in `form.scss`

PR #47, opened 2021-01-25
Cy tests refactor

EGiataganas:
### Description of change

Refactor cypress tests and split them into multiple contexts with different tests to improve our test readability


siapankina (approved):
Well done! :shipit:

PR #50, opened 2021-01-27
Add Caller name and Contact number / switch to use ScheduleRepair HACT endpoint

benbaumann95:
https://trello.com/c/v2DonIwK/307-as-an-agent-i-can-add-the-contact-name-and-number-to-a-work-order-so-that-the-contractor-can-contact-the-resident

### DIFF
--- a/src/utils/hact/raise-repair-form.js
+++ b/src/utils/hact/raise-repair-form.js
@@ -1,10 +1,16 @@
+import { v4 as uuidv4 } from 'uuid'
 import {
   mapPriorityCodeToHact,
   calculateRequiredCompletionDateTime,
 } from './priority-codes'
 
-export const buildRaiseRepairFormData = (formData) => {
+export const buildScheduleRepairFormData = (formData) => {
   return {
+    reference: [
+      {
+        id: uuidv4(),
+      },
+    ],
     descriptionOfWork: formData.descriptionOfWork,
     priority: {
       priorityCode:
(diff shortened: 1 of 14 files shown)


src/utils/hact/raise-repair-form.js
```diff
@@ -31,17 +37,59 @@ export const buildRaiseRepairFormData = (formData) => {
...
+          reference: [
+            {
+              id: formData.propertyReference,
+            },
+          ],
         },
+      ],
+    },
+    instructedBy: {
+      name: 'Hackney Housing',
```
benbaumann95, line 56: Defaulting to this to now.. It's not being used but is a required field

src/utils/hact/raise-repair-form.js
```diff
@@ -31,17 +37,59 @@ export const buildRaiseRepairFormData = (formData) => {
...
+    },
+    assignedToPrimary: {
+      name: `Contractor ${formData.sorCodeContractorRef}`,
```
benbaumann95, line 59: This will have the name of the contractor when this exists in the api but for now sending up the reference code

src/utils/hact/raise-repair-form.js
```diff
@@ -31,17 +37,59 @@ export const buildRaiseRepairFormData = (formData) => {
...
         ],
       },
-    ],
+    },
+    customer: {
+      name: formData.callerName,
+      person: {
+        name: {
+          full: formData.callerName,
+        },
```
benbaumann95, line 73: Not sure if this is the right thing here.. 
The customer could be different to the contact details that the operative provides on the raise repair form. Would this info come from the tenancy/resident API instead ?

NeilvB (approved):

EGiataganas (approved):

PR #51, opened 2021-01-27
Simulate basic contractor login and route their homepage to jobs list

NeilvB:
### Description of change

Support simulated contractor logins by way of a **temporary API route which sets a cookie**. This interim solution has been added so we can start testing the contractor pages / journeys. This temporary route and cookie logic will be deleted once we're certain we can onboard external contractors to our auth provider (Google). This functionality around cookies has been intentionally left untested because this logic will soon be replaced.

Includes an [extension to our current user object](https://github.com/LBHackney-IT/repairs-hub-frontend/commit/d835b78adcb5d6a541a3e6b0231687267cc684ec#diff-5f647c5cc86db387e2cdd6341b459488e331a03210f757e959fd6a03fd153cd3R21) that we use on both the backend in our auth checks and also as part of our context on the client. 

The user object now includes permission convenience properties and methods (e.g. `hasAgentPermission`) which we will be able to use for displaying different content to users in our pages. The function which builds the user is responsible for keeping a mapping between the group names we get from Google, and the internal roles we want in our application.

Note that this PR does not contain any differentiation between the different groups of contractors - they are all just contractors for now. This will need extending for the various types of contractors, who must only be able to access their own jobs.

In addition, this **does not include any blocking of user access of pages, whether regular pages or API routes**. This has been left out of this PR because we need to verify exactly which resources each role can access in the system. However, the new user object introduced here should help support that later work.

Differentiating between contractors and blocking roles from accessing particular resources, if needed, will be added in later PRs.

### Story Link

https://trello.com/c/8sP1ul0h/128-needs-estimating-as-a-contractor-admin-i-can-log-into-my-work-area-so-that-i-can-pick-up-the-jobs

### Before Merge

- [x] Notify team about new environment variables required for local development
- [x] Add `...contractors-alphatrack-group` to AWS Parameter Store to be picked up by serverless


.env.sample
```diff
@@ -8,11 +8,14 @@ REPAIRS_SERVICE_API_KEY=repairsApiKey
...
 HACKNEY_JWT_SECRET=sekret
-HACKNEY_AUTHORISED_GROUP=repairs-hub-frontend-staging
+REPAIRS_AGENTS_GOOGLE_GROUPNAME=repairs-hub-frontend-staging
+CONTRACTORS_GOOGLE_GROUPNAME=repairs-hub-frontend-staging-contractors-alphatrack
 
 #Redirect URL to run code locally
 REDIRECT_URL=localhost:5000
 
 #E2E
 CYPRESS_GSSO_TEST_KEY=
```
EGiataganas, line 18: How about deleting this since is not used anymore?

src/utils/user.test.js
```diff
@@ -0,0 +1,82 @@
+import { buildUser, AGENT_ROLE, CONTRACTOR_ROLE } from './user'
+
+const {
+  REPAIRS_AGENTS_GOOGLE_GROUPNAME,
+  CONTRACTORS_GOOGLE_GROUPNAME,
+} = process.env
+
+describe('buildUser', () => {
```
benbaumann95, line 8: nice tests 👍

benbaumann95 (approved):

.env.sample
NeilvB: Thanks. Will do this separately.

serverless.yml
```diff
@@ -38,7 +38,8 @@ functions:
       GSSO_URL: ${ssm:/repairs-hub/${self:provider.stage}/gsso-url}
       GSSO_TOKEN_NAME: ${ssm:/repairs-hub/${self:provider.stage}/gsso-token-name}
       HACKNEY_JWT_SECRET: ${ssm:/repairs-hub/${self:provider.stage}/hackney-jwt-secret}
-      HACKNEY_AUTHORISED_GROUP: ${ssm:/repairs-hub/${self:provider.stage}/authorised-group}
+      REPAIRS_AGENTS_GOOGLE_GROUPNAME: ${ssm:/repairs-hub/${self:provider.stage}/authorised-group}
```
EGiataganas, line 41: Another tiny change, which I am happy if it will be done separately is maybe renaming the key `authorised-group` in parameter stores in `agents-authorised-group` maybe. This will give us a clearer understanding of the group's role

EGiataganas (approved):

serverless.yml
NeilvB: Thanks. Noted.

PR #54, opened 2021-02-01
Allow adding multiple SOR codes to raise a repair

benbaumann95:
### Description of change
- Permit to add multiple SOR code select components (code and quantity) and enable deletion if there is more than one of these components
- Adds error validation to targeted components based on index
- Populate hidden fields of description and contractor ref once SOR code is selected
- Return array of RateScheduleItems on HACT raise repair form
- Show sor code description next to code in dropdown select (remove SOR code summary auto populate feature)

### Story Link

https://trello.com/c/fLciBa9w/263-as-an-agent-i-can-add-multiple-sor-codes-to-a-work-order-so-that-i-can-raise-all-the-repairs-required

### **Single SOR code**
Screenshot 2021-02-01 at 15 28 36

### **Multiple SOR code**
Screenshot 2021-02-01 at 15 27 35

### **Error validations targeting SOR component at right index**
Screenshot 2021-02-01 at 15 28 23

### **Mobile view**
Screenshot 2021-02-01 at 14 31 34

### DIFF
--- a/cypress/integration/raise-repair/form.spec.js
+++ b/cypress/integration/raise-repair/form.spec.js
@@ -2,16 +2,13 @@
 
 import 'cypress-audit/commands'
 
-beforeEach(() => {
-  cy.loginWithAgentRole()
-
-  cy.server()
-  // Stub request for raise a repair form page
-  cy.fixture('schedule-of-rates/codes.json').as('sorCodes')
-})
-
-describe('Raise repair form on property with tenure', () => {
+describe('Raise repair form', () => {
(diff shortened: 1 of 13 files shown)


NeilvB (approved):
Looks really good. Just a couple of `console.log` I think we might be able to lose from [this function](https://github.com/LBHackney-IT/repairs-hub-frontend/blob/1fb0f8bf22923c87ba2caf5846dc8122303cb488/src/components/Property/RaiseRepair/RaiseRepairForm.js#L27)

PR #59, opened 2021-02-04
Add tasks and sors tab to work order view page

benbaumann95:
### Description of change
- Add tasks and sors tab to work order view page
- Refactor the way we display the tabs using a Tabs component
- Supply the required tabs inside TabsList array and Tabs index component should autogenerate the html for this
- Add govuk js

Screenshot 2021-02-05 at 09 03 36

### DIFF
--- a/src/utils/date.js
+++ b/src/utils/date.js
@@ -29,3 +29,9 @@ export const convertToDateFormat = (string) => {
   const dateAsString = `${string.date}T${string.time}`
   return convertDate(dateAsString)
 }
+
+export const sortedByDateAdded = (data) => {
+  data.forEach((job) => (job.dateAdded = convertDate(job.dateAdded)))
+
+  return data.sort((a, b) => b.dateAdded - a.dateAdded)
+}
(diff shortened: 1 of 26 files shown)


src/utils/date.js
```diff
@@ -29,3 +29,9 @@ export const convertToDateFormat = (string) => {
   const dateAsString = `${string.date}T${string.time}`
   return convertDate(dateAsString)
 }
+
+export const sortedByDateAdded = (data) => {
+  data.forEach((job) => (job.dateAdded = convertDate(job.dateAdded)))
+
+  return data.sort((a, b) => b.dateAdded - a.dateAdded)
+}
```
NeilvB, line 37: Looks like there's an opportunity for a mini-refactor here, as both this and the `sortedByDate` function above do similar work.

How about something like this to replace both? (Untested)

```
export const sortObjectsByDateKey = (objects, dateKey) => {
  objects.forEach((o) => { o[dateKey] = convertDate(o[dateKey]) })

  return objects.sort((a, b) => b[dateKey] - a[dateKey])
}
```

Which could be called like this:

```
sortObjectsByDateKey(jobs, 'dateAdded')

// OR

sortObjectsByDateKey(jobs, 'dateRaised')
```

Later this could be extended to accept a sorting direction.

NeilvB (approved):

PR #65, opened 2021-02-07
Update job

siapankina:
### Description of change

-Contractor can update jobs
-can add new SOR code
-original SOR code is unchangeable
-reused SorCodeSelectView (with some updates)

### Story Link

https://trello.com/c/f9d9MPHl/290-as-a-contractor-admin-i-need-to-be-able-to-add-more-sors-to-raised-jobs-aka-variation

Screenshot 2021-02-10 at 18 07 03
Screenshot 2021-02-10 at 18 07 14
Screenshot 2021-02-10 at 18 07 26
Screenshot 2021-02-10 at 18 07 39

### DIFF
--- a/cypress/integration/jobs/update-job.spec.js
+++ b/cypress/integration/jobs/update-job.spec.js
@@ -1,0 +1,206 @@
+/// <reference types="cypress" />
+import 'cypress-audit/commands'
+
+describe('Contractor update a job', () => {
+  beforeEach(() => {
+    cy.loginWithContractorRole()
+    cy.server()
+    cy.fixture('schedule-of-rates/codes.json').as('sorCodes')
+    cy.fixture('work_orders/jobs.json').as('workorderslist')
+    cy.fixture('work_orders/tasks.json').as('tasksList')
+    cy.route('GET', 'api/repairs/?PageSize=10&PageNumber=1', '@workorderslist')
+    cy.route('GET', 'api/repairs/00012346/tasks', '@tasksList').as(
+      'taskListRequest'
+    )
+    cy.route('GET', 'api/schedule-of-rates/codes', '@sorCodes').as(
+      'sorCodeRequest'
+    )
+
+    cy.route({
+      method: 'POST',
+      url: '/api/jobStatusUpdate',
+      response: '',
+    }).as('apiCheck')
+  })
+
+  it('throws errors if input values are empty or not valid', () => {
+    cy.visit(`${Cypress.env('HOST')}/`)
+
+    cy.get('.govuk-table__cell').within(() => {
+      cy.contains('00012346')
+      cy.contains('a', 'Update').click()
+    })
+    cy.contains('Update work order: 00012346')
+    cy.get('form').within(() => {
+      cy.get('[type="radio"]').check('Update')
+      cy.get('[type="submit"]').contains('Next').click()
+    })
+
+    cy.wait('@taskListRequest')
+    cy.wait('@sorCodeRequest')
+
+    // Update page
+    cy.contains('Update work order: 00012346')
+
+    cy.contains('+ Add another SOR code').click()
+
+    cy.get('form').within(() => {
+      cy.get('[type="submit"]').contains('Next').click()
+    })
+
+    cy.get('form').within(() => {
+      cy.contains('Please select an SOR code')
+      cy.contains('Please enter a quantity')
+    })
+
+    cy.get('#repair-request-form').within(() => {
+      // Select SOR Code from dropdown
+      cy.get('select[id="sorCodesCollection[0][code]"]').select(
+        'DES5R004 - Emergency call out'
+      )
+      // Enter a non-number quantity
+      cy.get('input[id="sorCodesCollection[0][quantity]"]').clear().type('x')
+
+      cy.get('[type="submit"]').contains('Next').click()
+
+      cy.contains('Quantity must be a whole number')
+      // Enter a quantity more than the maximum
+      cy.get('input[id="sorCodesCollection[0][quantity]"]').clear().type('60')
+
+      cy.contains('Quantity must be 50 or less')
+      // Enter a quantity less then the minimum
+      cy.get('input[id="sorCodesCollection[0][quantity]"]').clear().type('0')
+      cy.contains('Quantity must be 1 or more')
+
+      // Enter a non-integer quantity
+      cy.get('input[id="sorCodesCollection[0][quantity]"]').clear().type('1.5')
+      cy.contains('Quantity must be a whole number')
+    })
+  })
+
+  it('allows the user to update the job by changing the existing quantity', () => {
+    cy.visit(`${Cypress.env('HOST')}/`)
+    cy.get('.govuk-table__cell').within(() => {
+      cy.contains('00012346')
+      cy.contains('a', 'Update').click()
+    })
+    cy.contains('Update work order: 00012346')
+    cy.get('form').within(() => {
+      cy.get('[type="radio"]').check('Update')
+      cy.get('[type="submit"]').contains('Next').click()
+    })
+    cy.wait('@taskListRequest')
+    cy.wait('@sorCodeRequest')
+    cy.get('#repair-request-form').within(() => {
+      cy.get('#quantity-0-form-group').within(() => {
+        cy.get('input[id="quantity-0"]').clear().type('0')
+      })
+      cy.get('[type="submit"]').contains('Next').click()
+    })
+    cy.get('[type="submit"]').contains('Confirm and close').click()
+    cy.get('@apiCheck')
+      .its('request.body')
+      .should('deep.equal', {
+        relatedWorkOrderReference: {
+          id: '00012346',
+          description: '',
+          allocatedBy: '',
+        },
+        typeCode: 8,
+        moreSpecificSORCode: {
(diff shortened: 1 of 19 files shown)


cypress/integration/jobs/update-job.spec.js
```diff
@@ -0,0 +1,165 @@
+/// <reference types="cypress" />
+import 'cypress-audit/commands'
+
+const TimeToWaitForRenderToFinish = 5000
```
benbaumann95, line 4: Does adding a default max wait time to cypress.json not work?

cypress/integration/raise-repair/form.spec.js
```diff
@@ -19,7 +19,6 @@ describe('Raise repair form', () => {
       response: '10102030',
     }).as('apiCheck')
   })
-
```
benbaumann95, line 22: remove space by accident? 😄

cypress/integration/jobs/update-job.spec.js
siapankina: unfortunately not , because the problem is not time out, the problem is detached from DOM

src/components/Property/RaiseRepair/SorCodeSelect.js
```diff
@@ -20,9 +22,16 @@ const SorCodeSelect = ({
         <Select
           name={`sorCodesCollection[${index}][code]`}
           label="SOR Code"
-          options={sorCodesList}
+          options={sorCodesList.map((e) => {
+            return {
+              value: e,
+              text: e,
+            }
+          })}
```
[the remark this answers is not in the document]
siapankina: forgot to delete it, was testing something 
good catch

src/components/Property/RaiseRepair/SorCodeSelect.js
```diff
@@ -20,9 +22,16 @@ const SorCodeSelect = ({
...
+          defaultValue={code}
           onChange={(event) => onSorCodeSelect(index, event)}
           required={true}
+          selected={code}
```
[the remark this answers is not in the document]
siapankina: not really, will update it

.eslintrc
```diff
@@ -25,6 +25,7 @@
   "rules": {
     "linebreak-style": ["error", "unix"],
     "require-atomic-updates": "off",
+    "cypress/no-unnecessary-waiting": "off",
```
NeilvB, line 28: Do we still need this default override now we're using `wait` on the requests rather than the time? I think we should stick to the defaults if we don't need to change this after all.

src/components/WorkOrders/SummaryUpdateJob.js
```diff
@@ -0,0 +1,70 @@
...
+
+  const tableData = (data) => {
+    return data.map((t, index) => (
+      <tr className={`task-${index} govuk-table__row`}>
+        <th scope="row" className={` sor-code-${index} govuk-table__header`}>
+          {t.code}
+        </th>
+        <td className={`quantity-${index} govuk-table__cell`}>{t.quantity}</td>
+        <td className={`edit-${index} govuk-table__cell`}>
+          <a onClick={changeStep}>Edit</a>
```
NeilvB, line 22: I think we might need an `href="#"` on this link, otherwise the user doesn't get the normal link cursor when they hover.

src/utils/hact/update-job.test.js
```diff
@@ -0,0 +1,42 @@
+import { buildUpdateJob } from './update-job'
```
NeilvB, line 1: Nice test

src/utils/update-job.js
```diff
@@ -0,0 +1,12 @@
+export const updatedTasks = (e) => {
+  let numberOfTasks = e.sorCodesCollection
+    ? (Object.keys(e).length - 1) / 2
+    : Object.keys(e).length / 2
```
NeilvB, line 4: How come we're dividing by two here? I was a bit confused why that's needed.

cypress/integration/jobs/update-job.spec.js
```diff
@@ -0,0 +1,164 @@
...
+      cy.get('input[id="quantity-0"]').clear().type('10')
+    })
+    cy.get('#repair-request-form').within(() => {
+      cy.get('.repairs-hub-link').click()
+      // Select SOR Code from dropdown
+      cy.get('select[id="sorCodesCollection[0][code]"]').select(
+        'DES5R004 - Emergency call out'
+      )
+
+      // Enter a non-number quantity
```
NeilvB, line 114: I don't think this comment applies for this input, please could we remove if so

cypress/integration/jobs/update-job.spec.js
```diff
@@ -0,0 +1,164 @@
...
+    )
+
+    cy.route({
+      method: 'POST',
+      url: '/api/jobStatusUpdate',
+      response: '',
+    }).as('apiCheck')
+  })
+
+  it('trhoughs errors if input values are empty or not valid', () => {
```
benbaumann95, line 26: Typo. Think it should be `throws`

src/components/WorkOrders/UpdateJob.js
```diff
@@ -0,0 +1,145 @@
...
+    )
+    makePostRequest(updateJobFormData)
+  }
+
+  const makePostRequest = async (formData) => {
+    setLoading(true)
+
+    try {
+      await postUpdateJob(formData)
+      router.push('/')
```
benbaumann95, line 60: what is this doing ?

src/utils/hact/update-job.js
```diff
@@ -0,0 +1,27 @@
...
+      allocatedBy: 'string',
+    },
+    typeCode: 8,
+    moreSpecificSORCode: {
+      rateScheduleItem: allTasks.map((task) => {
+        let codeAndName = task.code.split(' - ')
+        let showDescription =
+          task.description &&
+          task.description.length > 0 &&
+          task.description !== 'string'
```
benbaumann95, line 16: A bit confused of what we are trying to do here

src/utils/hact/update-job.js
```diff
@@ -0,0 +1,27 @@
+export const buildUpdateJob = (existingTasks, addedTasks, reference) => {
+  const allTasks = [...existingTasks, ...addedTasks]
+  return {
+    relatedWorkOrderReference: {
+      id: reference,
+      description: 'string',
+      allocatedBy: 'string',
```
benbaumann95, line 7: Do we need to send this up ?

src/utils/update-job.js
benbaumann95: I'm a bit confused with what is happening here. Maybe worth adding a comment if it's not possible to simplify?

src/components/WorkOrders/SummaryUpdateJob.js
siapankina: I will add `href="#" ` to SummaryCloseJob as well then
good catch ;)

src/utils/update-job.js
siapankina: to be able to get info on existing tasks
they are saved in formData separately 
`{sor-code-0: "DES5R004 - Emergency call out", quantity-0: 1, sor-code-1: "PLP5R054 - RENEW MIXER TAP", quantity-1: 1, sorCodesCollection: Array(1)}`
sorCodeCollection: Array is all SOR code that were added extra

siapankina: Actually I think there is a better way for that

benbaumann95: Might be worth having the existing tasks managed in its own collection array as there may be a case where it is multiple sors

src/components/WorkOrders/UpdateJob.js
siapankina: it redirects contractors to work-orders list table

src/utils/hact/update-job.js
siapankina: by mistake I added value `"string"` to some tasks, so some might come with description `"string"`

src/components/WorkOrders/UpdateJobForm.js
```diff
@@ -0,0 +1,110 @@
...
+                  name={`sor-code-${index}`}
+                  label={index == 0 ? 'SOR code' : ''}
+                  widthClass={`sor-code-${index} govuk-!-width-full`}
+                  register={register}
+                  readOnly={true}
+                  value={`${t.code} ${
+                    t.description && t.description !== 'string'
+                      ? '- ' + t.description
+                      : ''
+                  }`}
```
benbaumann95, line 54: Bit confused here. What is this doing?

siapankina: ### regarding "string" the same as before some will come with description = "string" 
and  existing tasks are a bit inconsistent, some come as code and description in `code` field  and some with code in code field and description in `description` field

src/utils/frontend-api-client/tasks.js
```diff
@@ -0,0 +1,7 @@
+import axios from 'axios'
+
+export const getTask = async (reference) => {
```
NeilvB, line 3: Please could we rename this to `getTasks`.

NeilvB (approved):

benbaumann95 (approved):

EGiataganas (approved):

PR #67, opened 2021-02-08
Repairs job pagination

EGiataganas:
### Description of change

Repairs job pagination

### Story Link

https://trello.com/c/P2wTKLnv

### DIFF
--- a/src/components/WorkOrders/JobView.js
+++ b/src/components/WorkOrders/JobView.js
@@ -35,13 +35,26 @@ const JobView = () => {
     setLoading(true)
     workOrderView()
   }, [])
+
+  const handlePageClick = (pageNumber) => {
+    setPageNumber(pageNumber)
+    workOrderView(pageNumber)
+    setLoading(true)
+  }
+
   return (
     <>
       {loading ? (
         <Spinner />
       ) : (
         <>
-          {workOrders && <JobsTable workOrders={workOrders} />}
+          {workOrders && (
+            <JobsTable
+              workOrders={workOrders}
+              pageNumber={pageNumber}
+              handlePageClick={handlePageClick}
+            />
+          )}
           {error && <ErrorMessage label={error} />}
         </>
       )}
--- a/src/components/WorkOrders/JobView.js
+++ b/src/components/WorkOrders/JobView.js
@@ -6,15 +6,16 @@ import ErrorMessage from '../Errors/ErrorMessage/ErrorMessage'
 import { sortObjectsByDateKey } from '../../utils/date'
 
 const JobView = () => {
+  const [pageNumber, setPageNumber] = useState(1)
   const [workOrders, setWorkOrders] = useState([])
   const [error, setError] = useState()
   const [loading, setLoading] = useState(false)
 
-  const workOrderView = async () => {
+  const workOrderView = async (pageNumber) => {
     setError(null)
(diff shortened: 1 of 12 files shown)


src/components/WorkOrders/JobView.js
```diff
@@ -6,15 +6,16 @@ import ErrorMessage from '../Errors/ErrorMessage/ErrorMessage'
 import { sortObjectsByDateKey } from '../../utils/date'
 
 const JobView = () => {
+  const [pageNumber, setPageNumber] = useState(1)
```
benbaumann95, line 9: Should we not have page number within the url so it's easily shareable and a user can track what page they are on ? At the moment the url stays constant right so if someone would refresh the page they would just start off at page 1 again

benbaumann95: Think you can use Router here perhaps

src/components/WorkOrders/JobsTable.js
```diff
@@ -38,6 +38,26 @@ const JobsTable = ({ workOrders }) => (
...
+      {pageNumber && (
+        <button
+          className="govuk-button next-page-button"
+          data-module="govuk-button"
+          onClick={() => handlePageClick(pageNumber + 1)}
+        >
+          Next page
+        </button>
+      )}
+    </div>
```
benbaumann95, line 60: I'd have a look on https://hmcts-design-system.herokuapp.com/components/pagination about design for pagination. Not sure a button is the right thing to use - maybe can check in with a designer

src/utils/frontend-api-client/repairs.js
```diff
@@ -1,7 +1,11 @@
 import axios from 'axios'
 
-export const getRepairs = async () => {
-  const { data } = await axios.get('/api/repairs')
+const PAGE_SIZE = 10
+
+export const getRepairs = async (pageNumber = 1) => {
+  const { data } = await axios.get(
+    `/api/repairs/?PageSize=${PAGE_SIZE}&PageNumber=${pageNumber}`
```
NeilvB, line 7: We could supply the `PageSize` and `PageNumber` with the `params` key (https://github.com/axios/axios#request-config) which would make this a bit easier to read.

e.g. (untested)

```
await axios.get(
  '/api/repairs/',
  { 
    params:  {
      PageSize: PAGE_SIZE,
      PageNumber: pageNumber
    }
  }
)
```

src/components/WorkOrders/JobsTable.js
EGiataganas: For now, we agreed that it will be a button next and previous only

src/components/WorkOrders/__snapshots__/JobsTable.test.js.snap
```diff
@@ -132,6 +132,9 @@ exports[`JobsTable component should render properly 1`] = `
         </tr>
       </tbody>
     </table>
+    <div
+      class="page-navigation govuk-!-padding-bottom-5"
+    />
```
benbaumann95, line 137: You need to add `pageNumber` and `handlePageClick` props to `JobsTable.test` otherwise the buttons won't appear here in the screenshot

benbaumann95 (approved):
Just needs an integration spec and a comment about supplying props for the snapshot

PR #73, opened 2021-02-11
Cancel work order

benbaumann95:
### Description of change

- Posts to the completeWorkOrder endpoint with typeCode 0 (other) and otherType 'cancel'
- Notes appear as `Cancelled: ${note}`
- 1 hour window from raising a repair to being able to cancel

### Story Link

https://trello.com/c/fgsIVYrC/264-as-an-rcc-agent-i-need-to-be-able-to-cancel-the-work-order-if-information-changes-or-i-make-a-mistake

### **Link to cancel work order**
Screenshot 2021-02-11 at 14 00 00

### **Cancel work order form / page**
Screenshot 2021-02-11 at 14 00 07

### **Repair cancelled confirmation screen**
Screenshot 2021-02-11 at 14 00 48

### **Status updated to work cancelled**
Screenshot 2021-02-11 at 14 00 30

### **Notes from cancellation**
Screenshot 2021-02-11 at 14 00 42


cypress/integration/cancel-repair/cancel-work-order.spec.js
```diff
@@ -0,0 +1,152 @@
+/// <reference types="cypress" />
+
+import 'cypress-audit/commands'
+
+describe('Cancel work order', () => {
```
NeilvB, line 5: I like this test.

One approach we could take for a readability bonus is to use a Date object we control and refer to within the test. I was reading a bit about cypress fixtures and JS dates. Looks like we might be able to show the reader exactly what date we're comparing to here and what the difference in minutes should be for the boundaries of our tests. It makes use of `.then` in cypress (https://docs.cypress.io/api/commands/fixture.html#Modifying-fixture-data-before-using-it).

This is a proposed idea in pseudocode:

```
  const workOrderRaisedTime = new Date()

  beforeEach(() => {
    //...

      cy.fixture('repairs/work-order.json')
      .then((workOrder) => {
        workOrder.dateRaised = workOrderRaisedTime.toISOString()  // override fixture property
      })
      .as('workOrder')
      // ...
  })

// ...
      
   it('Cancel a work order', () => {
        const cancelTime = new Date(workOrderRaisedTime)

        cancelTime.setMinutes(workOrderRaisedTime.getMinutes() + 59)

        cy.clock(cancelTime)

        // ... etc 
```
Might need an update for the check we do on the request object (e.g. `eventTime` we send up).

And a similar approach for the following test, except maybe using this 

` cancelTime.setMinutes(workOrderRaisedTime.getMinutes() + 61)`

NeilvB (approved):
Looks great. I've left a couple of non-blocking ideas which I'm keen to hear your thoughts on.

cypress/integration/cancel-repair/cancel-work-order.spec.js
benbaumann95: Yes, I agree this would improve readability and convey what the tests are trying to achieve a bit better here 👍

PR #78, opened 2021-02-15
Search by work order reference

EGiataganas:
### Description of change

Search by work order reference

### Story Link

https://trello.com/c/TdiQ7AES

### Screenshot

Screenshot 2021-02-16 at 12 17 40

Screenshot 2021-02-17 at 09 38 47


src/components/Search/Search.js
```diff
@@ -12,11 +12,16 @@ const Search = ({ query }) => {
   const [loading, setLoading] = useState(false)
   const [error, setError] = useState()
   const router = useRouter()
+  const patt = /[0-9]{1}[0-9]{7}/g
```
NeilvB, line 15: Please could we use a more specific name like `workOrderReferenceRegex`.

Does `/[0-9]{8}/g` give us the same result with simpler code?

cypress/integration/search_for_property.spec.js
```diff
@@ -2,66 +2,147 @@
 
 import 'cypress-audit/commands'
 
-describe('Search for property', () => {
+describe('Search by work order reference, postcode or address', () => {
```
NeilvB, line 5: Please update the name of this spec file to include work orders, or maybe better, use a separate file for work order searches.

src/components/Search/Search.js
```diff
@@ -41,8 +46,12 @@ const Search = ({ query }) => {
   const handleSubmit = (e) => {
     e.preventDefault()
 
-    propertiesURL(searchQuery)
-    searchQuery && searchForProperties(searchQuery)
+    if (patt.test(searchQuery)) {
+      workOrderUrl(searchQuery)
```
NeilvB, line 50: At the moment we send the user to a work order URL if the query matches our test. But what if they enter an 8 digit number which does not refer to any existing work order e.g. 00000000? I think we'd need something to stop them seeing a 404 in that case. Might be worth checking out what the old app did here.

src/components/Search/Search.js
EGiataganas: Thanks for pointing this out.
First I had `/[0-9]{8}/g` to match any 8 digit number but then I thought of change it to `/[1]{1}[0-9]{7}/g` to match the current work number ref. Although it seems I lost it on a stash 😂 
I will change it to `/[0-9]{8}/g`so it match any 8 digit number.

cypress/integration/search_for_property.spec.js
EGiataganas: I was thinking the same thing, I think I will just rename the spec file, as splitting it on a separate file for work order searches means we will need to set up another login and start the server which I believe will slow down our specs

benbaumann95 (commented):
Hey, i'd change the home text from `Find property` to `Find repair job or property` (from v1) now that we can search for a work order too

src/components/Search/Search.js
benbaumann95: Yeah I agree here. If only React had some `rescue_from` functionality like ruby does. Worth an investigation

src/components/WorkOrder/WorkOrderView.js
```diff
@@ -33,9 +33,16 @@ const WorkOrderView = ({ workOrderReference }) => {
       setWorkOrder(null)
       setProperty(null)
       console.log('An error has occured:', e.response)
-      setError(
-        `Oops an error occurred with error status: ${e.response?.status}`
-      )
+
+      if (e.response?.status === 404) {
```
benbaumann95, line 37: This looks nice. I think we could also redirect them back to the home page right so they can perform another search? At the moment they would need to navigate back to the home page to do this if I understand correctly

benbaumann95: I'm thinking display the home page first and then show this message at the top

src/components/Search/Search.js
```diff
@@ -7,29 +7,30 @@ import ErrorMessage from '../Errors/ErrorMessage/ErrorMessage'
...
 const Search = ({ query }) => {
-  const [searchParams, setSearchParams] = useState('')
+  const [searchQuery, setSearchQuery] = useState('')
   const [properties, setProperties] = useState([])
   const [loading, setLoading] = useState(false)
-  const [searching, setSearching] = useState(true)
-  const [searchQuery, setSearchQuery] = useState('')
   const [error, setError] = useState()
   const router = useRouter()
+  const patt = /[0-9]{8}/g
```
benbaumann95, line 15: Could we rename this to something a bit more descriptive? 
i.e.`workOrderReferencePattern` or I think Neil may have mentioned `workOrderReferenceRegex`

benbaumann95: As a foreign user to the system I'm not sure they would know what this meant at initial glance

benbaumann95 (approved):

src/components/Search/Search.js
```diff
@@ -39,63 +40,49 @@ const Search = ({ query }) => {
...
-            <hr className="govuk-section-break govuk-section-break--l govuk-section-break--visible" />
-            <p className="govuk-heading-s">
-              We found 0 matching results for: {decodeURI(searchQuery)}
-            </p>
-          </div>
-        </>
-      )
-    }
+  const workOrderUrl = (searchQuery) => {
+    router.push(`/work-orders/${searchQuery}`)
```
siapankina, line 67: 👍

siapankina (approved):
🚢

PR #79, opened 2021-02-15
Don’t rely on transitive dependencies

EGiataganas:
### Description of change

Don’t rely on transitive dependencies

For example, if React uses PropTypes internally (without exposing them), it could potentially remove them or change their major version, and do so in a patch release. This is safe if they’re not exposed. Therefore we should never rely on other packages’ transitive dependencies just “being there”. So let's explicitly add them to package.json to avoid them breaking.


package.json
```diff
@@ -62,6 +62,7 @@
     "lint-staged": "^10.5.3",
     "node-mocks-http": "^1.10.0",
     "prettier": "^2.2.1",
+    "prop-types": "^15.7.2",
```
NeilvB, line 65: I wonder if this needs to be in `dependencies`, see https://www.npmjs.com/package/prop-types#how-to-depend-on-this-package

PR #81, opened 2021-02-16
Control page access by role

NeilvB:
### Description of change

Restrict access to existing pages for only those roles we know require them.

| Feature | Permit agent? | Permit contractor? |
| ------------- | ------------- | --------------|
|  Search for property / work order  | Yes | No |
| View property | Yes  | No |
| Raise Repair | Yes  | No |
| View Repair  | Yes  | No |
| Repair List  | No  | Yes* |
| Close Repair  | No  | Yes* |
| Cancel Repair  | Yes | No |
| Vary Repair  | No  | Yes* |
| New Appointment (placeholder page)  | Yes  | No |

*contractor access scoped to their own specific repairs and related child resources is required, but will be controlled at the service API level and is not part of this PR.

### Story Link

https://trello.com/c/8m3seX6B/379-restrict-page-permissions


__test__/index.test.js
```diff
@@ -0,0 +1,10 @@
+import Home from '../src/pages/index'
+import { AGENT_ROLE, CONTRACTOR_ROLE } from '../src/utils/user'
+
+describe('Home.permittedRoles', () => {
+  ;[AGENT_ROLE, CONTRACTOR_ROLE].forEach((role) => {
+    it(`permits the ${role} role to access the page`, () => {
+      expect(Home.permittedRoles).toContain(role)
+    })
+  })
+})
```
benbaumann95, line 10: I can see in our Home component that we are conditionally rendering what component an agent/contractor should be redirected to. Do we need to go down a layer and be more specific i.e. make sure that only contractor can only see <JobView />. Or perhaps even separate it out as AgentHome and ContractorHome components etc. 

I think your other specs cover this scenario actually but just thinking out loud here

cypress/integration/raise-repair/form.spec.js
```diff
@@ -378,3 +378,22 @@ describe('Raise repair form', () => {
     cy.audit()
   })
 })
+
+// This is a "special" integration test to end-to-end test
```
benbaumann95, line 382: 👍

benbaumann95 (approved):
Hey Neil - I love how you've separated each permission and spec into their own commit. Made this review very clear and concise.
I'm also a big fan of how you've future proofed these tests (reminds me of the pundit specs from northstar 😉 ) to allow more roles to be easily added without any significant code restructure.
Just left a minor comment - not blocking merging at all

🎉

__test__/index.test.js
NeilvB: Thanks for the comment. I believe we should be OK with just setting permissions on all the page components, as long as we have integration specs which go down each conditional in cases like the home page. 

I think having permissions on lower level components might get tricky because the permissions matching might then need to drill down into each child component to check against the user role. I think that since all routes start in Next JS with a page component I think we should be OK in combination with continuing our integration spec coverage.

What I have done is [update the home page spec with a context for the contractor log in](https://github.com/LBHackney-IT/repairs-hub-frontend/pull/81/commits/67bfda7e6423a9846f137dcfb78126d5747bee13?diff=unified&w=1). This was taken from a separate spec but I think having them together makes it easier for us to be sure about the logic working on the home page.

Interestingly, I think @EGiataganas raised the idea of having separate home components for agents and contractors in a separate discussion. I don't have a strong preference from my side either way about that at the moment though.

PR #83, opened 2021-02-17
Show last 50 records in repairs history, changed from a default of 10

benbaumann95:
### Description of change

- Repairs endpoint only returns a default of 10. Until we decide on pagination / display a link to load more repairs - this is a temporary fix to load 50 of the last records

### DIFF
--- a/src/utils/frontend-api-client/repairs.js
+++ b/src/utils/frontend-api-client/repairs.js
@@ -1,11 +1,12 @@
 import axios from 'axios'
 
-const PAGE_SIZE = 10
+const PAGE_SIZE_CONTRACTORS = 10
+const PAGE_SIZE_AGENTS = 50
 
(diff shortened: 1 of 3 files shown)


src/utils/frontend-api-client/repairs.js
```diff
@@ -1,26 +1,28 @@
 import axios from 'axios'
 
-const PAGE_SIZE = 10
+const PAGE_SIZE_CONTRACTORS = 10
+const PAGE_SIZE_AGENTS = 50
 
 export const getRepairs = async (pageNumber = 1) => {
-  const { data } = await await axios.get('/api/repairs/', {
```
benbaumann95, line 6: Would have thought prettier/linter would have picked this up!

NeilvB (approved):

PR #86, opened 2021-02-19
Filter trade contractor sor

benbaumann95:
### Description of change

- Includes https://github.com/LBHackney-IT/repairs-hub-frontend/pull/77 (needs to be shipped as well due to the upcoming api breaking change where UH sor codes and priorities are migrated)
- Moves endpoint calls into more organised file naming convention - to match the api 
- Add a DataList component to use for Trade select
- Add trade -> contractor -> sor code filter logic where each select field is disabled until the former has been selected
- Remove duplicate fixtures
- Hardcode (temporarily) the SOR codes response at the vary job stage (upcoming ticket https://trello.com/c/MuO8leHn/372-as-a-contractor-admin-i-can-type-in-the-sor-code-so-that-i-can-vary-the-job needs will address this)

### Story Link

https://trello.com/c/a11J58Oh/322-as-a-user-i-can-select-from-a-list-of-sor-codes-filtered-by-trade
https://trello.com/c/fLciBa9w/263-as-an-agent-i-can-add-multiple-sor-codes-to-a-work-order-so-that-i-can-raise-all-the-repairs-required


src/utils/frontend-api-client/contractors.js
```diff
@@ -0,0 +1,9 @@
+import axios from 'axios'
+
+export const getContractors = async (propertyReference, tradeCode) => {
+  const { data } = await axios.get(
```
NeilvB: I've got a slight preference for using the params object for axios rather than a string with interpolated query params. One of the reasons is that we get to split it out over multiple lines which could be useful for git diffing.

This could be rewritten

```
  ...axios.get('/api/contractors', {
      params: {
        propertyReference,
        tradeCode
      }
  })
```

NeilvB (commented):
Looking great. Lots to juggle in this story. Just the temporary deploy commit to remove IMO.

NeilvB (approved):

PR #89, opened 2021-02-23
Show contacts on raise repair form

NeilvB:
### Description of change

Add resident contact information to the base of the raise repairs form.

Redact sensitive information for responses from the Node API which are not for agents as they have no need to see this.

 ### Before Merge

 - [x] Double-check with backend if JSON structure they return is correct - should contacts be a nested object?

### Story Link

https://trello.com/c/ckaBKNzB/345-as-an-rcc-agent-i-can-view-the-contact-name-and-number-assigned-to-a-property-when-raising-a-repair-so-i-can-check-details-if-ne

### Screenshots

These are from the base of the raise repairs form.

image

image

### DIFF
--- a/cypress/fixtures/properties/property.json
+++ b/cypress/fixtures/properties/property.json
@@ -42,18 +42,16 @@
     "typeCode": "SEC",
     "typeDescription": "Secure"
   },
-  "contacts": {
-    "contacts": [
-      {
-        "firstName": "Mark",
-        "lastName": "Gardner",
-        "phoneNumbers": ["00000111111", "00000222222"]
-      },
-      {
-        "firstName": "Luam",
-        "lastName": "Berhane",
-        "phoneNumbers": ["00000666666"]
-      }
-    ]
-  }
+  "contacts": [
+    {
+      "firstName": "Mark",
+      "lastName": "Gardner",
+      "phoneNumbers": ["00000111111", "00000222222"]
+    },
+    {
+      "firstName": "Luam",
+      "lastName": "Berhane",
+      "phoneNumbers": ["00000666666"]
+    }
+  ]
 }
--- a/cypress/fixtures/properties/property.json
+++ b/cypress/fixtures/properties/property.json
@@ -41,5 +41,19 @@
   "tenure": {
     "typeCode": "SEC",
     "typeDescription": "Secure"
+  },
+  "contacts": {
+    "contacts": [
+      {
+        "firstName": "Mark",
+        "lastName": "Gardner",
+        "phoneNumbers": ["00000111111", "00000222222"]
+      },
+      {
+        "firstName": "Luam",
+        "lastName": "Berhane",
+        "phoneNumbers": ["00000666666"]
+      }
+    ]
   }
 }
--- a/cypress/integration/raise-repair/form.spec.js
+++ b/cypress/integration/raise-repair/form.spec.js
@@ -13,6 +13,7 @@ describe('Raise repair form', () => {
     cy.fixture('schedule-of-rates/trades.json').as('trades')
     cy.fixture('contractors/contractors.json').as('contractors')
     cy.fixture('properties/property.json').as('property')
+
     cy.route('GET', 'api/properties/00012345', '@property')
     cy.route(
       'GET',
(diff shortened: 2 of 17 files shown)


src/components/Property/Contacts/ContactsNoneAvailable.js
```diff
@@ -0,0 +1,15 @@
+const ContactsNoneAvailable = () => {
```
benbaumann95, line 1: Wonder if we can make this component more reusable such as `WarningText` and then pass the string value as props. Seems like nothing is custom apart from the text.

The reason I raise this is that one of the stories coming up will need to use the same warning text template. But happy to address this and refactor when the time comes

src/components/Property/Contacts/ContactsTable.js
```diff
@@ -0,0 +1,37 @@
+import PropTypes from 'prop-types'
+import ContactsRow from './ContactsRow'
+
+const ContactsTable = ({ contacts }) => {
+  return (
+    <table className="govuk-table">
```
benbaumann95, line 6: You may have missed it but Andrew has added some reusable <Table> <Thead> <TBody> components etc. Going forward might be worth using this (so we don't have to repeat all the generic classnames) but I know older table components don't have this updated yet so could address as one PR later

cypress/fixtures/properties/property.json
```diff
@@ -41,5 +41,19 @@
   "tenure": {
     "typeCode": "SEC",
     "typeDescription": "Secure"
+  },
+  "contacts": {
```
benbaumann95, line 45: I've seen your comments to propose a change to this nesting and totally agree that this makes no sense like this

cypress/integration/raise-repair/form.spec.js
```diff
@@ -55,6 +56,14 @@ describe('Raise repair form', () => {
       ]
     )
 
+    // Property contact information table
+    cy.get('.govuk-table__caption').contains('Contacts')
+    cy.get('.govuk-table__row .govuk-table__cell').contains('Mark Gardner')
+    cy.get('.govuk-table__row .govuk-table__cell').contains('00000111111')
+    cy.get('.govuk-table__row .govuk-table__cell').contains('00000222222')
+    cy.get('.govuk-table__row .govuk-table__cell').contains('Luam Berhane')
+    cy.get('.govuk-table__row .govuk-table__cell').contains('00000666666')
```
benbaumann95, line 65: Is it possible to add a data-attribute to identify the rows exactly? Maybe a bit overkill but I find it useful when doing tests on multiple records to make sure i'm targeting the correct index/row etc

benbaumann95 (approved):
Left some minor thoughts but not blocking merging

cypress/fixtures/properties/property.json
NeilvB: Backend has been updated, and frontend matched up here: https://github.com/LBHackney-IT/repairs-hub-frontend/pull/89/commits/499c4f19c4bdf645eb77c6893bbee9a5e72a5bfb

src/components/Property/Contacts/ContactsTable.js
NeilvB: Thanks. I'm going to leave this for now and I think we can address as a sweep later.

cypress/integration/raise-repair/form.spec.js
NeilvB: Thanks. I definitely agree that these should be more focused. I was interested to read about the Cypress recommendation to use data attributes just for testing.

For this case I think I'm mostly interested in the numbered row in the table and I hoped to just rely on the DOM ordering. I've chosen to go for using `cy.eq()` to scope the query to the first and then the second row. I feel a little resistant to adding data attributes to our markup if they don't support the functioning or presentation of the app, but I can also see how that would be super useful for testing too especially as we're using lots of generic classes.

I'm going to go ahead with this for now as it is an improvement hopefully, but happy and interested to talk about this further.

```
    cy.get('.govuk-table')
      .contains('Contacts')
      .parent()     // grab the parent table for scoping the following
      .within(() => {
        cy.get('tbody>tr').eq(0).contains('Mark Gardner')   // first row
        cy.get('tbody>tr').eq(0).contains('00000111111')
        cy.get('tbody>tr').eq(0).contains('00000222222')

        cy.get('tbody>tr').eq(1).contains('Luam Berhane')  // second row
        cy.get('tbody>tr').eq(1).contains('00000666666')
      })
```

src/components/Property/Contacts/ContactsNoneAvailable.js
NeilvB: Great shout. I've extracted a `WarningText` here: https://github.com/LBHackney-IT/repairs-hub-frontend/pull/89/commits/aa696b0ae5ed1cee3dcfe9dd037f37fbdc078c81

PR #90, opened 2021-02-23
Add load more button

EGiataganas:
### Description of change

Add load more button to load an additional 50 repairs in the history list, so that I can see the full history

### Story Link

https://trello.com/c/4qxLSdb5

### DIFF
--- a/cypress/integration/show_property.spec.js
+++ b/cypress/integration/show_property.spec.js
@@ -54,7 +54,7 @@ describe('Show property', () => {
         )
       })
 
-      it('Display the repairs history for the property', () => {
+      it('Displays the repairs history for the property', () => {
         // Repairs history tab should be active
         cy.get('.govuk-tabs__list-item--selected a').contains('Repairs history')
 
@@ -104,7 +104,7 @@ describe('Show property', () => {
         cy.audit()
       })
 
-      it('Display no repairs text when records do not exist', () => {
+      it('Displays no repairs text when records do not exist', () => {
         cy.route(
           'GET',
           'api/repairs/?propertyReference=00012345&PageSize=50&PageNumber=1',
@@ -162,7 +162,7 @@ describe('Show property', () => {
     }
   )
 
-  context('Display property with no tenure type', () => {
+  context('Displays property with no tenure type', () => {
     beforeEach(() => {
       // Stub request with property response
       cy.fixture('properties/property_no_tenure.json').as('property')
--- a/cypress/integration/show_property.spec.js
+++ b/cypress/integration/show_property.spec.js
@@ -18,7 +18,7 @@ describe('Show property', () => {
         cy.fixture('repairs/work-orders.json').as('workOrders')
         cy.route(
           'GET',
-          'api/repairs/?propertyReference=00012345&PageSize=50',
+          'api/repairs/?propertyReference=00012345&PageSize=50&PageNumber=1',
           '@workOrders'
         )
         cy.visit(`${Cypress.env('HOST')}/properties/00012345`)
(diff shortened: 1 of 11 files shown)


cypress/integration/show_property.spec.js
```diff
@@ -54,13 +54,77 @@ describe('Show property', () => {
...
+
+        it('Displays the load more button', () => {
+          cy.get('.page-navigation').within(() => {
+            cy.contains('Load more')
+          })
+          // Run lighthouse audit for accessibility report
+          cy.audit()
+        })
+
+        it('Displays more repairs after clicking Load more button', () => {
```
NeilvB, line 116: I like this test.

But can we delete the test above it? It looks to me like this does everything that one does if we add a wrapping function around our clicking of 'Load more'

```
    cy.get('.page-navigation').within(() => {
      cy.contains('Load more').click()
    })
```

cypress/integration/show_property.spec.js
```diff
@@ -54,13 +54,77 @@ describe('Show property', () => {
...
+            'api/repairs/?propertyReference=00012345&PageSize=50&PageNumber=2',
+            '@moreWorkOrders'
+          )
 
-        cy.get('.govuk-tabs').within(() => {
-          cy.get('.govuk-tabs__tab').contains('Repairs history')
-          cy.get('.govuk-heading-l').contains('Repairs history')
+          // Click load more button
+          cy.contains('Load more').click()
+          // cy.visit(`${Cypress.env('HOST')}/properties/00012345`)
```
NeilvB, line 127: This commented code can be ✂️

cypress/integration/show_property.spec.js
```diff
@@ -70,44 +134,30 @@ describe('Show property', () => {
             cy.contains('th', 'Description')
           })
```
NeilvB, line 135: Should we also expect the existing properties to be shown in this test? At the moment it could seem like this replaces the existing properties but I think our requirement is to add more to an existing list. If this test is getting long as a result, we could split out a separate test, e.g. `cypress/integration/property_show/repair_history_tab.spec.js` or similar

src/components/Property/RepairsHistory/RepairsHistoryView.js
```diff
@@ -7,19 +7,27 @@ import { getRepairsForProperty } from '../../../utils/frontend-api-client/repair
 import { sortObjectsByDateKey } from '../../../utils/date'
 
 const RepairsHistoryView = ({ propertyReference, tabName }) => {
+  const [pageNumber, setPageNumber] = useState(1)
   const [workOrders, setWorkOrders] = useState([])
+  const [arrayForHoldingworkOrders, setArrayForHoldingworkOrders] = useState([])
```
benbaumann95, line 12: Think this should be `arrayForHoldingWorkOrders`

benbaumann95: Can't we just call it `AllWorkOrders` ?

src/components/Property/RepairsHistory/RepairsHistoryView.js
```diff
@@ -33,13 +41,28 @@ const RepairsHistoryView = ({ propertyReference, tabName }) => {
 
   useEffect(() => {
     setLoading(true)
+    setPageNumber(pageNumber + 1)
 
-    getRepairsHistoryView(propertyReference)
+    getRepairsHistoryView(propertyReference, pageNumber)
   }, [])
 
+  const handleShowMoreClick = (newPageNumber) => {
```
benbaumann95, line 49: Is there a better name for this perhaps? i.e. `handleLoadMoreWorkOrders` or even `loadMoreWorkOrders` ?
What do you reckon.

src/components/Property/RepairsHistory/RepairsHistoryView.js
EGiataganas: I would go with
```
const [workOrdersPerPage, setWorkOrdersPerPage] = useState([]) // which will hold work orders per page
const [workOrders, setWorkOrders] = useState([]) // which will hold all work orders
```

NeilvB (approved):

PR #93, opened 2021-02-24
Type in sor

benbaumann95:
### Description of change

**Enter full SOR code into text input at job vary stage for contractors**
- Only those within a contractor group can perform the action to request the SOR code id endpoint. The contractor ref in the JWT header is used to determine which contractor is making the request and used as a param (behind the scenes in the API) along with the propertyReference. Upon entering a code, an onBlur event handler will trigger and return a 404 if no valid SOR code is found or if  there is a matching SOR code then the promise will resolve and return that particular record along with its description and cost details. (locally you will need to update the x-hackney-user to a JWT with contractor permissions)
- Short description is populated below the SOR code text input and cost saved as a hidden field to display on the summary page
- Total cost calculated on summary page and displayed under the tasks table
- State is a bit tricky with firing off async requests and causing re-renders (causing values to wipe from form fields) so useState is handling the form object values
- We send back up the unique id (guid) for existing rate schedule items to signal to  the API that these are in fact original items (there is a validation in the API to check this matches and if not will return an error). We don't send up the id column or value for the additional rate schedule items and the unique id's for this are generated within the API

- PR also includes refactoring and a slight reorganising of folders/files - including renaming of components to align more with HACT standard i.e. RateScheduleItem to describe the items associated with SOR codes (i.e. quantity, cost, description etc.)
- Adds missing snapshot tests for existing components

### Story Link

https://trello.com/c/MuO8leHn/372-as-a-contractor-admin-i-can-type-in-the-sor-code-so-that-i-can-vary-the-job

### Known issues [OPTIONAL]

With multiple RateScheduleItem components, every time an API request is made, the entire screen loads (values are unchanged though). Could be improved by adding a spinner to the unique id's of each RateScheduleItem component but can be addressed in a design ticket down the line

**Error after a 404 returns from the API**
Screenshot 2021-02-24 at 12 28 17

**Error when trying to submit the form**
Screenshot 2021-02-24 at 12 28 42

**Summary page with cost breakdown**
Screenshot 2021-02-24 at 13 45 31


siapankina:
well done 👍

siapankina (approved):

PR #102, opened 2021-03-10
Show original tasks separately within tasks and sors tab

benbaumann95:
### Description of change

Show original and varied tasks separately within the tasks and sors tab on a work order
Screenshot 2021-03-10 at 16 57 27

### Story Link

https://trello.com/c/CXGGjfao/22-as-a-contract-manager-i-can-see-a-breakdown-of-original-and-varied-sor-codes-and-quantities-so-i-know-what-im-approving

### DIFF
--- a/src/components/WorkOrder/TasksAndSors/TasksAndSorsRow.js
+++ b/src/components/WorkOrder/TasksAndSors/TasksAndSorsRow.js
@@ -1,35 +1,31 @@
 import PropTypes from 'prop-types'
 import { formatDateTime } from '../../../utils/time'
+import { TR, TD } from '../../Layout/Table'
 
 const tasksAndSorsRow = ({
   code,
   description,
   dateAdded,
-  quantity,
+  taskQuantity,
   cost,
   status,
   index,
 }) => (
-  <tr
-    data-row-id={index}
-    className="govuk-table__row govuk-table__row--clickable govuk-body-s"
-  >
-    <td className="govuk-table__cell">{code}</td>
-    <td className="govuk-table__cell">{description}</td>
-    <td className="govuk-table__cell">
-      {dateAdded ? formatDateTime(dateAdded) : '—'}
-    </td>
-    <td className="govuk-table__cell">{quantity}</td>
-    <td className="govuk-table__cell">{cost}</td>
-    <td className="govuk-table__cell">{status}</td>
-  </tr>
+  <TR index={index} className="govuk-body-s">
+    <TD>{code}</TD>
+    <TD>{description}</TD>
+    <TD>{dateAdded ? formatDateTime(dateAdded) : '—'}</TD>
+    <TD>{taskQuantity}</TD>
+    <TD>£{cost}</TD>
+    <TD>{status}</TD>
+  </TR>
 )
 
 tasksAndSorsRow.propTypes = {
   code: PropTypes.string.isRequired,
   description: PropTypes.string.isRequired,
   dateAdded: PropTypes.instanceOf(Date).isRequired,
-  quantity: PropTypes.number.isRequired,
+  taskQuantity: PropTypes.number.isRequired,
   cost: PropTypes.number.isRequired,
   status: PropTypes.string.isRequired,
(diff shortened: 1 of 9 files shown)


src/components/WorkOrder/TasksAndSors/TasksAndSorsView.js
```diff
@@ -44,7 +46,11 @@ const TasksAndSorsView = ({ workOrderReference, tabName }) => {
       ) : (
         <>
           {tasksAndSors && (
-            <TasksAndSorsTable tasksAndSors={tasksAndSors} tabName={tabName} />
+            <TasksAndSorsTable
+              tasksAndSors={tasksAndSors}
```
NeilvB, line 50: Not a big deal but I wonder if it's worth giving this property and variable a more specific name now that we have `originalTasksAndSors` to avoid potential confusion later. Does it make sense for this to be `postVariationTasksAndSors` perhaps?

src/components/WorkOrder/TasksAndSors/TasksAndSorsRow.js
```diff
@@ -1,28 +1,26 @@
...
-    </td>
-    <td className="govuk-table__cell">{quantity}</td>
-    <td className="govuk-table__cell">{cost}</td>
-    <td className="govuk-table__cell">{status}</td>
-  </tr>
+  <TR index={index} className="govuk-body-s">
+    <TD>{code}</TD>
+    <TD>{description}</TD>
+    <TD>{dateAdded ? formatDateTime(dateAdded) : '—'}</TD>
+    <TD>{isOriginal ? originalQuantity : quantity}</TD>
```
NeilvB, line 20: Is there a way to pass in a different `quantity` from a parent component if this is an original SOR? Just thinking it might be neat to surface the logic higher up, and then this component wouldn't need to know/care about whether the SOR is original or not. Just an idea.

NeilvB (approved):

siapankina (approved):

src/components/WorkOrder/TasksAndSors/TasksAndSorsView.js
benbaumann95: I think this can look exactly the same as `originalTasksAndSors` if no variation was done with it. How about `currentTasksAndSors` or something to refer to it being the latest snapshot of the live tasks and sors against the work order

src/components/WorkOrder/TasksAndSors/TasksAndSorsRow.js
benbaumann95: Good idea - i'll take a look thanks

benbaumann95: Hey so this is actually coming from 
https://github.com/LBHackney-IT/repairs-hub-frontend/blob/main/src/components/WorkOrder/TasksAndSors/TasksAndSorsTable.js#L33
which is an attribute on the actual object so i'm not setting anything in the parent component for this quantity

NeilvB: I suppose we _could_ still manage that by overriding properties on the item from the parent. I kind of like the idea of the row only knowing about one quantity. But this feels really nitpicky and subjective so happy either way.

Not tested:

```
          {tasks.map((entry, index) => (
            <TasksAndSorsRow
              key={index}
              index={index}
              {...entry, quantity: isOriginal ? entry.originalQuantity : entry.quantity }
            />
          ))}
```

benbaumann95: ooh I like it

benbaumann95: This syntax doesn't work. I've done some reading and you can't do destructuring and reasssign attributes that easily. Would involve some more steps. Going to just send it down separately

NeilvB: Ah :( in that case I'm sorry for the unnecessary detour

PR #111, opened 2021-03-22
Allow partial match when deciding slot choice text

NeilvB:
### Description of change

Previously the frontend would do a match on fairly specific text when deciding how to present slots. This commit makes that matching broader so the frontend will work with some slight variations in slot descriptions.


siapankina (approved):
😎

PR #113, opened 2021-03-23
Spend limit messages

NeilvB:


benbaumann95 (approved):
Approving but just leaving a reminder that you were going to add the `hub-user` call to fit in with the other API requests upon page load

PR #124, opened 2021-03-29
Add contract managers

EGiataganas:
### Description of change

- Add new `contract manager` google group
- Contract manager will have the same permissions as the contractor and agent
- Contract manager will be redirected to the same dashboard as the contractor with work orders with status code 90 only

### Story Link

https://trello.com/c/RaKeUK2Q

### DIFF
--- a/src/components/WorkOrders/JobView.js
+++ b/src/components/WorkOrders/JobView.js
@@ -20,7 +20,7 @@ const JobView = ({ pageNumber }) => {
     setError(null)
 
     try {
-      var data = []
+      let data = []
 
       if (user.hasContractorPermissions) {
         data = await getRepairs(pageNumber)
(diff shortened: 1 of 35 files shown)


src/components/WorkOrders/JobView.js
```diff
@@ -15,7 +20,13 @@ const JobView = ({ pageNumber }) => {
...
 
     try {
-      const data = await getRepairs(pageNumber)
+      var data = []
+
+      if (user.hasContractorPermissions) {
+        data = await getRepairs(pageNumber)
+      } else if (user.hasContractManagerPermissions) {
+        data = await getPendingApprovalRepairs(pageNumber)
+      }
```
benbaumann95, line 29: Probs should use `let` for ES6 syntax

Could you not do `const data = user.hasContractorPermissions ? await getRepairs(pageNumber) : await getPendingApprovalRepairs(pageNumber)`

benbaumann95 (approved):
Just left a minor comment

src/components/WorkOrders/JobView.js
EGiataganas: Sure happy to use `let` so that it follows ES6 syntax.

In addition, since we also have `hasAgentPermissions` I thought it's better to be more strict.

PR #125, opened 2021-03-29
Add confirmation screen for job variation 

benbaumann95:
### Description of change

- Add update confirmation screens when contractor varies a job
- When authorisation is required, present red banner with message
- When authorisation is not required, present green banner with message
- Add links to view work order and jobs dashboard on this page

- Update confirmation panel colours to use latest design system colours https://design-system.hackney.gov.uk/developing/colours/

Screenshot 2021-03-29 at 16 31 46
Screenshot 2021-03-29 at 16 44 04

### DIFF
--- a/src/utils/time.js
+++ b/src/utils/time.js
@@ -1,11 +1,15 @@
 const DAY = 86400000
 
-export const extractTimeFromDate = (date) =>
-  date.toLocaleString('en-GB', {
+export const extractTimeFromDate = (datetime) => {
+  if (typeof datetime === 'string') {
+    datetime = new Date(datetime)
+  }
+  return datetime.toLocaleString('en-GB', {
     hour: 'numeric',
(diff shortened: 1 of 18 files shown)


siapankina (approved):
😎

src/utils/time.js
```diff
@@ -1,11 +1,15 @@
 const DAY = 86400000
 
-export const extractTimeFromDate = (date) =>
-  date.toLocaleString('en-GB', {
+export const extractTimeFromDate = (datetime) => {
+  if (typeof datetime === 'string') {
+    datetime = new Date(datetime)
+  }
```
siapankina, line 6: just a suggestion.
maybe we can add a test to check` if` statement, pass string and check that it returns exactly what we want ?

EGiataganas (approved):

PR #141, opened 2021-04-19
Use priority code mappings consistent with backend

NeilvB:
When scheduling a repair we include the following priority fields:
```
priorityCode
priorityDescription
requiredCompletionDateTime
numberOfDays
```

This commit updates our priority code mappings, so we send back the same priority code as was returned from the API `/priorities` call.

It also updates some hour and day values for priorities.

By matching the backend DB `sor_priorities` table we should have consistent values for these

image

### DIFF
--- a/src/utils/hact/helpers/priority-codes.js
+++ b/src/utils/hact/helpers/priority-codes.js
@@ -1,27 +1,27 @@
-export const mapPriorityCodeToHact = {
+export const priorityCodeCompletionTimes = {
+  // immediate
   1: {
     numberOfHours: 2,
     numberOfDays: 0,
-    priorityCodeHact: 1,
   },
+  // emergency
   2: {
     numberOfHours: 24,
     numberOfDays: 1,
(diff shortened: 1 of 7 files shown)


src/utils/hact/helpers/priority-codes.js
```diff
@@ -1,27 +1,27 @@
-export const mapPriorityCodeToHact = {
+export const priorityCodeCompletionTimes = {
+  // immediate
   1: {
     numberOfHours: 2,
     numberOfDays: 0,
-    priorityCodeHact: 1,
   },
```
NeilvB, line 6: Matches https://github.com/LBHackney-IT/repairs-api-dotnet/blob/develop/RepairsApi/V2/Infrastructure/Migrations/20210118152301_SorCodeSeeding.cs#L48

siapankina (approved):
😎

benbaumann95 (approved):

PR #142, opened 2021-04-19
Add success page for variation approved/rejected

siapankina:
Created SuccessPage component so we can reuse it for contractors as well (might need to change some logic there, but we will be able to reuse it)

ticket: https://trello.com/c/DqmGK4x8/80-as-a-contract-manager-i-am-taken-to-a-confirmation-page-once-a-repair-has-been-approved-or-rejected-so-that-i-know-whats-happene
Screenshot 2021-04-20 at 15 22 02

### DIFF
--- a/src/components/SuccessPage/SuccessPage.js
+++ b/src/components/SuccessPage/SuccessPage.js
@@ -1,0 +1,41 @@
+import PropTypes from 'prop-types'
+import Link from 'next/link'
+
+const SuccessPage = ({ workOrderReference, text }) => {
+  return (
+    <div>
+      <section className="lbh-page-announcement">
+        <div className="lbh-announcement__content">
+          <p>
+            {text} <strong>work order {workOrderReference}</strong>
+          </p>
+        </div>
+      </section>
+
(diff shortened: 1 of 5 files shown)


src/components/SuccessPage/SuccessPage.js
```diff
@@ -0,0 +1,44 @@
+import PropTypes from 'prop-types'
+import Link from 'next/link'
+
+const SuccessPage = ({ workOrderReference, variationApproved }) => {
+  const text = variationApproved
```
benbaumann95, line 5: Should we just pass all this text as props?

So we can reuse this component for the RCC agent authorisation flow

benbaumann95 (approved):
Hey, great idea of making the success page a reusable component!
 Just left a thought about how we could make it even more reusable for future work

src/components/SuccessPage/SuccessPage.js
benbaumann95: I'm not sure this component needs to know whether a variation was approved or not - that can come from the parent. Then similar for when we have an authorisation approved

PR #143, opened 2021-04-20
Filter by status

benbaumann95:
### Description of change

https://trello.com/c/afUBxqh8/69-as-a-contractor-contract-manager-i-can-filter-by-status-on-my-dashboard-so-i-know-what-jobs-need-an-action-taken-by-me

- Filter work order by status
- Get available work order filters through /filters/WorkOrder endpoint
- Construct url query parameter as filter options are applied to allow sharing of urls with filters applied

**Collapsed filters**
Screenshot 2021-04-19 at 20 11 31

**Hidden filters**
Screenshot 2021-04-13 at 16 08 23


siapankina:
just a suggestion: 
what do you think about adding functionality` "clear all filters" `? 
so if contractor/ contract manager wants to see unfiltered list again they don't have to un-click all filtering options

benbaumann95:
> just a suggestion:
> what do you think about adding functionality`"clear all filters"`?
> so if contractor/ contract manager wants to see unfiltered list again they don't have to un-click all filtering options

Good idea but wasn't part of the A/C's so if it's necessary we can add a card to add this

PR #149, opened 2021-04-22
Fix contractor group detection for non-production environments

NeilvB:
[TRELLO CARD](https://trello.com/c/etSxCp8h/111-bug-fix-use-correct-contractor-group-detection-in-staging-and-development)

### Description of change

Correct detection of contractor groups in the staging and development environments.

Our group formats have changed so that `staging` is a suffix.

That's not a problem for our agent and contractor manager groups because those are described in a single environment variable.

But we currently do not correctly detect contractor groups because we ignore the staging suffix.

This PR adds a suffix for this and updates the regex for detection of this role.

Production should not have this variable set to anything so we can just rely on the fallback of no suffix.

### Before merge

 - [ ] Double check development parameter store variable for PREFIX
 - [ ] Add development parameter store variable for SUFFIX
 - [ ] Check staging parameter store variable for PREFIX
 - [ ] Add staging parameter store variable for SUFFIX
 - [ ] Tell dev team about updates required for local development

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -145,13 +145,13 @@ workflows:
             - e2e
           filters:
             branches:
-              only: develop
+              only: fix-contractor-group-detection
       - build-deploy-development:
           requires:
             - assume-role-development
           filters:
             branches:
-              only: develop
+              only: fix-contractor-group-detection
       - assume-role-staging:
           context: api-assume-role-staging-context
           requires:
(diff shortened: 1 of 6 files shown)


benbaumann95 (approved):

NeilvB:
Closing this for now because I need to add an update to detect production logins.

NeilvB:
Reopening now this has been updated to use a regex for flexibility.

PR #151, opened 2021-04-22
No DLO repairs appointments in hub

NeilvB:
### Description of change

Do not prompt agents raising DLO repairs of any priority to book an appointment.

This will (for the moment) be handled within DRS.

So instead we route them to the form success page with a link.

### Story Link

https://trello.com/c/olwGdpJ7/18-as-an-agent-i-see-the-work-confirmation-page-after-raising-any-dlo-order

### Before Merge

 - [x] Update Circle CI env var
 - [x] Update env var in dev, staging and production
 - [ ] Tell fellow devs about new env variable for local dev

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,7 +56,9 @@ jobs:
       - *attach_workspace
       - run:
           name: Run E2E With Lighthouse Audit
-          command: yarn e2e:audit
+          command: |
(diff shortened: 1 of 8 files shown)


benbaumann95 (approved):

siapankina (approved):
😎

PR #157, opened 2021-04-26
Add warning text

benbaumann95:
### Description of change

Display warning text if total cost of SOR code entries exceed raise limit (sum of unit cost  * quantity for all codes)

### Story Link

https://trello.com/c/9IQsL8ry/93-as-an-rcc-agent-i-can-see-a-warning-text-that-the-repair-is-over-my-spend-limit-so-that-i-know-it-will-be-sent-for-approval

Screenshot 2021-04-26 at 14 37 17

### DIFF
--- a/cypress/integration/raise-repair/form.spec.js
+++ b/cypress/integration/raise-repair/form.spec.js
@@ -13,7 +13,9 @@ describe('Raise repair form', () => {
     cy.fixture('schedule-of-rates/trades.json').as('trades')
     cy.fixture('contractors/contractors.json').as('contractors')
     cy.fixture('properties/property.json').as('property')
-    cy.route('GET', 'api/hub-user', {})
+    cy.fixture('hub-user/user.json').then((user) => {
+      cy.intercept('GET', 'api/hub-user', user)
+    })
 
     cy.route('GET', 'api/properties/00012345', '@property')
     cy.route(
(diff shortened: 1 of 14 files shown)


cypress/integration/raise-repair/form.spec.js
```diff
@@ -497,6 +499,73 @@ describe('Raise repair form', () => {
...
+
+    // Remove SOR at index 1 to make total cost 5.80 x 9 = 52.2
+    cy.get('button[id="remove-rate-schedule-item-1"]').click()
+    // Within user's raise limit so no warning text is displayed
+    cy.get('.govuk-warning-text.lbh-warning-text').should('not.exist')
+    // Edit quanitity to 43 to make total 5.80 x 43 = 249.4
+    cy.get('input[id="rateScheduleItems[2][quantity]"]').clear().type('43')
+    // Within user's raise limit so no warning text is displayed
+    cy.get('.govuk-warning-text.lbh-warning-text').should('not.exist')
+    // Edit quanitity to 43 to make total 5.80 x 44 = 255.2
```
NeilvB, line 560: should be 44 🔍

cypress/integration/raise-repair/form.spec.js
```diff
@@ -497,6 +499,73 @@ describe('Raise repair form', () => {
...
+    cy.get('input[id="rateScheduleItems[2][quantity]"]').clear().type('44')
+    // Warning text as user's raise limit (£250) has been exceeded
+    cy.get('.govuk-warning-text.lbh-warning-text').within(() => {
+      cy.contains(
+        'The works order cost exceeds the approved spending limit and will be sent to a manager for authorisation'
+      )
+    })
+  })
```
NeilvB, line 568: Nice test 🙌

src/components/Property/RaiseRepair/RaiseRepairForm.js
```diff
@@ -25,11 +26,15 @@ const RaiseRepairForm = ({
   trades,
   contacts,
   onFormSubmit,
+  currentUserRaiseLimit,
```
NeilvB, line 29: Wonder if we just call this `raiseLimit` as this component perhaps doesn't need to know where it's from...

src/components/Property/RaiseRepair/RateScheduleItemView.js
```diff
@@ -31,16 +34,55 @@ const RateScheduleItemView = ({
...
+      rateScheduleItemCostAtSameIndex.cost = costPerUnit
+      rateScheduleItemCostAtSameIndex.quantity = quantity
+
+      setRateScheduleItemCosts(rateScheduleItemCosts)
+      updateTotalCost(rateScheduleItemCosts)
+    } else {
+      const newRateScheduleItemCosts = [
+        ...rateScheduleItemCosts,
+        {
+          index: index,
```
NeilvB, line 61: Do we need an index here if we're using an array? Wondering if we can just index into it like 
`rateScheduleItemCosts[index] =  { cost: costPerUnit, quantity: quantity }`

NeilvB (approved):
Looks good. Left some very minor comment and a question about how we're using an array.

src/components/Property/RaiseRepair/RateScheduleItemView.js
NeilvB: If that's applicable I'm not sure exactly how that fits with our state hook method though..

benbaumann95: Ah this would be a lot simpler... Let me give it a go

benbaumann95: I think it becomes tricky when managing the state of adding and removing the ​RateScheduleItem components.
If i'm not keeping track of the index, then when using `slice()` to remove one of the elements from the array, the indexes change to be incremental again. But the way the multi components work is that there may be a RateScheduleItem with index 0 and index 4 with (1..3) removed. I'd need a way of targeting the index at 4 but this new array would just look like, for example:
```
0: {cost: 74.89, quantity: 3}
1: {cost: 24.89, quantity: 2}
```

Do you get what I mean - sorry if i'm not communicating this well. Hard to explain in words here - let me know if you want a better explanation. Happy if you wanted to fiddle around with the code yourself to make your simpler suggestion work when taking into consideration adding/removing SOR codes.

benbaumann95: I'm going to merge this for now (with all your other comments addressed) but I'd like to revisit this and make it work with your nicer suggestion above

PR #164, opened 2021-05-04
Show env name in header

siapankina:
Screenshot 2021-05-04 at 15 38 52


src/components/Layout/Header/Header.js
```diff
@@ -45,6 +45,11 @@ const HeaderComponent = ({ serviceName }) => {
                 </svg>
                 <span className="lbh-header__logo-text"> Hackney </span>
                 <span className="lbh-header__service-name">{serviceName}</span>
+                <span
```
NeilvB, line 48: Please can we add a conditional to only show this element if `process.env.NEXT_PUBLIC_ENV_NAME` is truthy?

That way, we won't render anything for production. Currently this would still render an empty span on prod I think.

NeilvB (approved):

PR #171, opened 2021-05-06
Do not show Authorisation Pending Approval filter for contractors

benbaumann95:
### Description of change

https://trello.com/c/2Jo9A9A7/147-as-a-contractor-i-cant-see-works-thats-are-pending-authorisation-so-i-dont-complete-the-work-until-its-approved

- When logged in as a contractor, the API scopes the work orders to not include those with Authorisation Pending Approval.
- We just need to hide the filter option on the frontend

### DIFF
--- a/src/components/WorkOrders/Filter/WorkOrdersFilter.js
+++ b/src/components/WorkOrders/Filter/WorkOrdersFilter.js
@@ -1,7 +1,10 @@
+import { useContext } from 'react'
 import PropTypes from 'prop-types'
 import Spinner from '../../Spinner/Spinner'
 import { Checkbox, Button } from '../../Form'
 import { GridColumn, GridRow } from '../../Layout/Grid'
+import UserContext from '../../UserContext/UserContext'
+import { STATUS_AUTHORISATION_PENDING_APPROVAL } from '../../../utils/status-codes'
 
 const WorkOrdersFilter = ({
   loading,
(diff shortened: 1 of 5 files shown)


src/components/WorkOrders/Filter/WorkOrdersFilter.test.js
```diff
@@ -27,22 +28,109 @@ describe('WorkOrdersFilter component', () => {
...
+          key: '1010',
+          description: 'Authorisation Pending Approval',
+        },
+        {
+          key: '90',
+          description: 'Variation Pending Approval',
+        },
+        {
+          key: '1010',
+          description: 'Authorisation Pending Approval',
```
NeilvB, line 41: Looks like a duplicate object in this array

src/components/WorkOrders/Filter/WorkOrdersFilter.js
```diff
@@ -10,6 +12,16 @@ const WorkOrdersFilter = ({
   appliedFilters,
   clearFilters,
 }) => {
+  const { user } = useContext(UserContext)
+
+  const statusFilterOptions = () => {
+    if (user && user.hasContractorPermissions) {
+      return filters.Status.filter((status) => status.key !== '1010')
```
NeilvB, line 19: I don't think it's necessarily something for this PR, but this makes me wonder if we should make a move to capture all of these status codes so we don't have to sometimes search around for their meaning - perhaps in environment variables? Or a config object with keys like `"authorisation_pending_approval"` mapped to the code and label text. That might make the views and cypress tests more readable and make things like typos less likely.

NeilvB (approved):

src/components/WorkOrders/Filter/WorkOrdersFilter.js
benbaumann95: Ah so I did create a `status-codes.js` file which i've forgotten to use here but yeah we should probably move to something a bit better like you suggest

PR #172, opened 2021-05-10
Respect operative working hours in target time calculations (part 2)

neilvanbeinum:
Update target time calculations to include some special rule for weekend raises.

Essentially an order raised on a Saturday should have its target calculated as if it had been raised on the preceding Friday.

This does not contain specific behaviour for other non-working days because the call centre is not open on a Sunday or bank holiday and we do not have rules for these.

Builds on the implementation from [part 1](https://github.com/LBHackney-IT/repairs-hub-frontend/pull/169)

[Trello card](https://trello.com/c/j5tgNsSd/21-respect-contractor-working-hours-when-calculating-target-dates)

### DIFF
--- a/src/utils/helpers/completionDateTimes.js
+++ b/src/utils/helpers/completionDateTimes.js
@@ -1,27 +1,25 @@
 import { priorityCodeCompletionTimes } from '../hact/helpers/priority-codes'
-import add from 'date-fns/add'
+import { isSaturday, addDays, subDays, isWeekend } from 'date-fns'
 
-const WEEKEND_DAYS = [0, 6]
+const isWorkingDay = (date) => !isWeekend(date)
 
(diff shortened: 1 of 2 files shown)


src/utils/helpers/completionDateTimes.js
```diff
@@ -1,14 +1,44 @@
...
+    return 0
+  }
+}
 
 const dateAfterCountWorkingDays = (startDate, targetWorkingDaysCount) => {
-  var workingDaysCount = 0
-  var calendarDaysCount = 0
+  // If raised on a weekend, include the remainder of that weekend for
+  // the count used to create the returned date
+  var startingOffsetDayCount = startingDayCountOffset(
```
benbaumann95, line 33: Think we could probably use ES6 `let` instead of `var` when variable re-assignment may happen.

PR #174, opened 2021-05-11
Schedule appointments using the DRS Web Booking Manager 

NeilvB:
Link to the Web Booking Manager following the raising of a DLO repair. This replaces the static DRS link with a link to a dedicated scheduling app.

[TRELLO CARD](https://trello.com/c/Heey60Xw/6-allow-agents-to-book-dlo-appointments-without-entering-drs-web-booking-manager-route)

The Web Booking Manager is linked from here:

image


src/components/Property/RaiseRepair/RaiseRepairFormView.js
```diff
@@ -40,15 +46,39 @@ const RaiseRepairFormView = ({ propertyReference }) => {
...
         setAuthorisationPendingApproval(true)
       } else if (externallyManagedAppointment) {
+        const cookies = new Cookies()
+
+        let schedulerSessionId = cookies.get(
+          process.env.NEXT_PUBLIC_DRS_SESSION_COOKIE_NAME
+        )
+
+        if (!schedulerSessionId) {
+          ;({ schedulerSessionId } = await getSchedulerSessionId())
```
benbaumann95, line 67: Curious what this syntax does here `;({ ... }`

benbaumann95 (approved):

src/components/Property/RaiseRepair/RaiseRepairFormView.js
NeilvB: This was a new one for me too. It's funky isn't it :) 

As you can see I'm using `let` to assign an initial value on line 63 and then have a need to reassign to that variable, and wanted to use an object destructure when doing so. Apparently JS can't interpret something like `{ myVar } = obj` as it will treat the curlies as a block. So wrapping in parens is a way around that. Then JS needs to also be told that we're not calling a function from a previous line which explains the `;`.

See the 'Assignment without declaration
' section and the note in particular here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#object_destructuring

benbaumann95: ooh that's neat - thanks for explaining that to me!

NeilvB: Once I added the parens, the linter took care of the `;` for me 🙂

PR #176, opened 2021-05-12
Authorisation summary 

siapankina:
ticket:
 https://trello.com/c/BtiMHma5/44-as-a-cm-i-can-see-a-summary-of-the-variation-request-before-submitting-it-so-i-can-clearly-see-the-details

Screenshot 2021-05-14 at 16 20 23

### DIFF
--- a/cypress/integration/work-order/variation-authorisation.spec.js
+++ b/cypress/integration/work-order/variation-authorisation.spec.js
@@ -4,25 +4,42 @@ import 'cypress-audit/commands'
 describe('Contract manager can authorise variation', () => {
   beforeEach(() => {
     cy.loginWithContractManagerRole()
-
     cy.server()
     // Stub requests
     cy.fixture('properties/property.json').as('property')
-    cy.route('GET', 'api/properties/00012345', '@property')
     cy.fixture('work-orders/status-variation-pending-approval.json').as(
       'workOrder'
     )
+    cy.fixture('work-orders/variation-tasks.json').as('variation-tasks')
+    cy.fixture('work-orders/tasks-and-sors.json').as('tasks-and-sors')
+
+    cy.route('GET', 'api/properties/00012345', '@property')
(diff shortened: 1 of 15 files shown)


EGiataganas (approved):

cypress/integration/work-order/variation-authorisation.spec.js
```diff
@@ -72,4 +96,87 @@ describe('Contract manager can authorise variation', () => {
...
+        'have.attr',
+        'href',
+        '/work-orders/10000012/variation-authorisation'
+      )
+      .click()
+
+    cy.contains('Authorisation variation request: 10000012')
+    cy.contains('Summary of Tasks and SORs')
+
+    cy.get('.origina-sor-summary').within(() => {
```
benbaumann95, line 114: small typo with class name

cypress/integration/work-order/variation-authorisation.spec.js
```diff
@@ -72,4 +96,87 @@ describe('Contract manager can authorise variation', () => {
...
+
+      //Unchanged task
+      cy.contains('td', 'Unchanged')
+      cy.contains('td', 'DES5R006')
+      cy.contains('p', 'Normal Call outs')
+      cy.contains('td', '£10')
+      cy.contains('td', '1')
+      cy.contains('td', '£10')
+      cy.contains('td', '1')
+      cy.contains('td', '£10')
```
benbaumann95, line 165: Could we possible scope this with data-attributes?
i.e. we should so something like `cy.get('[data-row-id="2"]').within(() => {..})` otherwise these matching values are too generic for the entire page

src/components/WorkOrder/Authorisation/VariationAuthorisationSummary.js
```diff
@@ -0,0 +1,220 @@
+import PropTypes from 'prop-types'
```
benbaumann95, line 1: Can we get a snapshot test for this ?

PR #178, opened 2021-05-17
Make notes an optional field when closing a work order

benbaumann95:
https://trello.com/c/gHIWveRw/141-make-adding-notes-on-work-order-complete-as-optional

### DIFF
--- a/src/utils/hact/work-order-complete/close-job.js
+++ b/src/utils/hact/work-order-complete/close-job.js
@@ -9,7 +9,7 @@ export const buildCloseWorkOrderData = (completionDate, notes, reference) => {
       {
         typeCode: 0,
         otherType: 'complete',
-        comments: notes,
+        comments: `Work order closed - ${notes}`,
         eventTime: completionDate,
       },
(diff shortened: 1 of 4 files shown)


siapankina (approved):

src/utils/hact/work-order-complete/close-job.js
```diff
@@ -9,7 +9,7 @@ export const buildCloseWorkOrderData = (completionDate, notes, reference) => {
       {
         typeCode: 0,
         otherType: 'complete',
-        comments: notes,
+        comments: `Work order closed - ${notes}`,
```
NeilvB, line 12: I noticed the format we use for cancellations is

`Cancelled: ${formData.cancelReason}`

Wonder if the user would get a little readability improvement if we keep them consistent? Just thinking about looking through a list of comments on a complex order. Maybe we could only conditionally show the `:` or `-` if there is a comment too..

NeilvB (approved):
Approved with minor string formatting comment

PR #184, opened 2021-05-18
Send up contractor ref as params for schedule-of-rates/codes/{sorCode} endpoint

benbaumann95:
- This helps determine the correct contract to use in terms of finding the relevant SOR the user is attempting to find
- Contractor may belong in one or more contractor groups so it is necessary to see what contract the current work order they are attempting to vary is associated with

### Story Link

https://trello.com/c/t64Y5RFU/196-send-up-contractor-ref-to-sorcode-id-endpoint-when-varying-a-work-order


src/utils/frontend-api-client/schedule-of-rates/codes.js
```diff
@@ -16,10 +16,15 @@ export const getSorCodes = async (
...
-export const getSorCode = async (code, propertyReference) => {
+export const getSorCode = async (
+  code,
+  propertyReference,
+  contractorReference = false
+) => {
   const { data } = await axios.get(`/api/schedule-of-rates/codes/${code}`, {
     params: {
       propertyReference: propertyReference,
+      ...(contractorReference && { contractorReference: contractorReference }),
```
NeilvB, line 27: I think we could skip specifying the property here as a micro-improvement

`...(contractorReference && { contractorReference }),`

NeilvB (commented):
Is it worth double checking the related integration spec to make sure we're not accidentally calling the real API instead of using our mocks? I wondered if we needed to add the contractor reference param to some mocks here and a few lines below to match the full request https://github.com/LBHackney-IT/repairs-hub-frontend/blob/develop/cypress/integration/work-order/update-work-order.spec.js#L40

src/utils/frontend-api-client/schedule-of-rates/codes.js
benbaumann95: Ahh nice syntax knowledge here 😉

benbaumann95:
> Is it worth double checking the related integration spec to make sure we're not accidentally calling the real API instead of using our mocks? I wondered if we needed to add the contractor reference param to some mocks here and a few lines below to match the full request https://github.com/LBHackney-IT/repairs-hub-frontend/blob/develop/cypress/integration/work-order/update-work-order.spec.js#L40

In the fixtures file `work-order.json` there is no contractor ref field so I think that's why this is passing OK without making an HTTP request. But I can add the column in because that's a better reflection of the work order object we get back in reality now so that will now require me to mock this ref too

NeilvB (approved):
👍

PR #185, opened 2021-05-18
Fixes for DRS Web Booking Manager session management

NeilvB:
Ensure session cookie deletion happens and set expiry to 12 hours.

Relates to [this Trello card](https://trello.com/c/Heey60Xw/6-allow-agents-to-book-dlo-appointments-without-entering-drs-web-booking-manager-route).

### DIFF
--- a/src/components/Property/RaiseRepair/RaiseRepairFormView.js
+++ b/src/components/Property/RaiseRepair/RaiseRepairFormView.js
@@ -42,6 +42,8 @@ const RaiseRepairFormView = ({ propertyReference }) => {
   const [currentUser, setCurrentUser] = useState()
   const router = useRouter()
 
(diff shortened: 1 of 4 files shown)


src/components/Property/RaiseRepair/RaiseRepairFormView.js
```diff
@@ -71,6 +71,7 @@ const RaiseRepairFormView = ({ propertyReference }) => {
             schedulerSessionId,
             {
               path: '/',
+              maxAge: 60 * 60 * 12, // 12 hours
```
EGiataganas, line 74: How about moving that calculation into a const variable at the top?

EGiataganas (approved):

PR #186, opened 2021-05-18
Drop ChooseOption and add MultiButton

EGiataganas:
### Description of change

Drop ChooseOption and add MultiButton component

### Story Link

https://trello.com/c/5EJa4lGg

### Screenshot
Screenshot 2021-05-18 at 18 37 54


src/components/WorkOrder/WorkOrderDetails.js
```diff
@@ -22,48 +20,39 @@ const WorkOrderDetails = ({
 }) => {
   const { user } = useContext(UserContext)
 
+  const allowedMenu = () => {
+    return WORK_ORDER_DROP_DOWN_MENU.filter((choice) => {
```
NeilvB, line 24: What do you think about renaming this module and constant to remove the 'drop down' and 'menu'? That describes a UI implementation which could change. The module on its own just describes the various actions and permissions on them. What do you think of 'workOrderActions' or 'workOrderChoices'?

src/components/WorkOrder/WorkOrderDetails.js
```diff
@@ -22,48 +20,39 @@ const WorkOrderDetails = ({
...
+            Works order: {workOrder.reference}
+          </h1>
+        </GridColumn>
+        <GridColumn width="one-third">
+          {allowedMenu()?.length > 0 && (
+            <MultiButton
+              name="workOrderMenu"
+              label="Select work order"
+              secondary={false}
+              choices={allowedMenu()}
```
NeilvB, line 50: We're calculating the allowed menu actions twice if the first passes the condition - once on this line and once on line 45. Can we get away with calculating them when the component loads once and then referring to that variable? E.g. assign `allowedMenu` to the result of the filtering instead of a function.

src/utils/workOrderDropDownMenu.js
```diff
@@ -0,0 +1,61 @@
...
+  AUTHORISATION_MANAGER_ROLE,
+} from './user'
+
+export const WORK_ORDER_DROP_DOWN_MENU = [
+  {
+    href: 'update',
+    title: 'Update',
+    description: 'Update Work Order',
+    permittedRoles: [CONTRACTOR_ROLE, CONTRACT_MANAGER_ROLE],
+    permittedStatus: WORK_ORDERS_STATUSES,
```
NeilvB, line 21: Nitpick but I think this should be plural - `permittedStatuses`

NeilvB (commented):
Looks good but I think we will benefit from the rename of the module I mention.

Very cool to have this set of actions cleaned up and to get a new component too.

NeilvB (approved):

PR #195, opened 2021-05-20
Fix auth manager repair raises

NeilvB:
Permit auth managers to raise repairs. 

This was implicitly being blocked by redacting the contact information for this user. 

It was done in such a way that the frontend couldn't interpret the API response. This PR also includes logic to reduce the chance of this falling over in the future.

### DIFF
--- a/src/components/Property/Contacts/Contacts.js
+++ b/src/components/Property/Contacts/Contacts.js
@@ -2,9 +2,19 @@ import PropTypes from 'prop-types'
 import WarningText from '../../Template/WarningText'
 import ContactsTable from './ContactsTable'
 
+const warningText = (contacts) => {
+  if (contacts.length < 1) {
+    return 'No contact details available for this property'
+  } else if (contacts[0] === 'REMOVED') {
+    return 'You are not permitted to view contact details'
+  }
+}
+
 const Contacts = ({ contacts }) => {
-  return contacts.length < 1 ? (
-    <WarningText text="No contact details available for this property" />
+  const text = warningText(contacts)
+
+  return text ? (
+    <WarningText text={text} />
   ) : (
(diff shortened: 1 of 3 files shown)


EGiataganas (approved):

src/components/Property/Contacts/Contacts.js
```diff
@@ -2,9 +2,19 @@ import PropTypes from 'prop-types'
 import WarningText from '../../Template/WarningText'
 import ContactsTable from './ContactsTable'
 
+const warningText = (contacts) => {
+  if (contacts.length < 1) {
+    return 'No contact details available for this property'
+  } else if (contacts[0] === 'REMOVED') {
+    return 'You are not permitted to view contact details'
```
benbaumann95, line 9: Do we need to let the end user know that they have no permission rather than just showing nothing.
Would they have expected to see contacts otherwise?

NeilvB: I guess I was coming at this from the perspective of us on the team adding a new role and having the same issue with the raise repair form. But this has made me realise this info should probably just be logged rather than actually shown in the UI. I'll have a bit more of a think about how to tackle this.

benbaumann95: Cool, I'm personally happy either way. 
Maybe could just merge so we can get the form working for the auth managers and then address this as a later PR perhaps

benbaumann95 (approved):

src/components/Property/Contacts/Contacts.js
NeilvB: OK. I'm going to merge as I think it's an improvement overall and I'm having a bit of trouble finding a better solution currently.

PR #200, opened 2021-05-21
Bank holiday aware target dates

neilvanbeinum:
Calculate completion times in the Node API and make use of an external API to fetch bank holiday data necessary for that.

Uses a simple cache in temporary lambda storage.

#### Before Merge

 - [x] Update development vars
 - [ ] Update staging vars
 - [ ] Update production vars
 - [ ] Tell dev team about new vars


src/utils/helpers/completionDateTimes.js
```diff
@@ -1,57 +1,79 @@
...
-  )
+export class PriorityCodeError extends Error {
+  constructor(message) {
+    super(message)
+  }
 }
 
-const isNonWorkingDay = (date) => isWeekend(date) || isBankHoliday(date)
+export class CompletionTimesCalculator {
+  constructor(priorityCode, bankHolidays = []) {
```
benbaumann95, line 11: Neat. Would love to chat about the use of constructor like this at some point

benbaumann95: Rails-esque

benbaumann95 (approved):
This looks awesome Neil, i've learnt a thing or two from reviewing this PR. 
Big fan of the way you've used classes to make the complicated business logic more manageable and readable here. Loving all the tests, particularly in `schedule.test.js`,  and the error handling class.

PR #216, opened 2021-06-04
Show various warning texts for DLO cancellations

NeilvB:
Conditionally show warning text for DLO cancellations.

[TRELLO CARD](https://trello.com/c/NsFRZYrh/24-add-note-for-dlo-cancellations)

image

### DIFF
--- a/src/utils/helpers/priorities.js
+++ b/src/utils/helpers/priorities.js
@@ -1,4 +1,9 @@
-// Currently, "normal" and "urgent" priority codes
-export const priorityCodesRequiringAppointments = [3, 4]
+export const IMMEDIATE_PRIORITY_CODE = 1
+export const EMERGENCY_PRIORITY_CODE = 2
+export const URGENT_PRIORITY_CODE = 3
+export const NORMAL_PRIORITY_CODE = 4
 
-export const highPriorityCodes = [1, 2]
+export const priorityCodesRequiringAppointments = [
+  URGENT_PRIORITY_CODE,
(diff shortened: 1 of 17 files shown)


src/utils/helpers/priorities.js
```diff
@@ -1,2 +1,9 @@
...
-export const priorityCodesRequiringAppointments = [3, 4]
+export const IMMEDIATE_PRIORITY_CODE = 1
+export const EMERGENCY_PRIORITY_CODE = 2
+export const URGENT_PRIORITY_CODE = 3
+export const NORMAL_PRIORITY_CODE = 4
+
+export const priorityCodesRequiringAppointments = [
+  URGENT_PRIORITY_CODE,
+  NORMAL_PRIORITY_CODE,
+]
```
EGiataganas, line 9: This looks very good. How about including the description as well? Maybe like that:

```
export const EMERGENCY_PRIORITY = { code: 2, description: '2 [E] EMERGENCY'}
```

By doing the above the `EMERGENCY_PRIORITY_CODE` will be EMERGENCY_PRIORITY.code, and you can also the hardcoded `priorityDescription`

NeilvB: Doesn't the backend supply the descriptions in the `GET
​/api​/v2​/schedule-of-rates​/priorities` endpoint? I'm not sure it benefits us to store it on the frontend too unless we have a particular reason to know about that.

EGiataganas (approved):

PR #217, opened 2021-06-07
Display TMO

siapankina:
ticket:  https://www.pivotaltracker.com/story/show/178315088

Screenshot 2021-06-07 at 18 07 54

### DIFF
--- a/src/components/Property/PropertyDetailsGrid.js
+++ b/src/components/Property/PropertyDetailsGrid.js
@@ -11,6 +11,7 @@ const PropertyDetailsGrid = ({
   tenure,
   hasLinkToProperty,
   canRaiseRepair,
+  tmoName,
 }) => {
   return (
     <div className="govuk-grid-row">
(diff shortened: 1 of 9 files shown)


src/components/Property/PropertyDetailsGrid.js
```diff
@@ -22,12 +23,12 @@ const PropertyDetailsGrid = ({
             subTypeDescription={subTypeDescription}
             hasLinkToProperty={hasLinkToProperty}
           />
-
           <TenureAlertDetails
```
NeilvB, line 26: Do you think this component might be better called just `TenureDetails` now? I wasn't sure if a TMO is technically an "alert" or just some additional info that might be useful to the viewer in certain circumstances

src/components/Property/TMO.js
```diff
@@ -0,0 +1,18 @@
+import PropTypes from 'prop-types'
+
+const TMO = ({ tmoName }) => {
```
NeilvB, line 3: TMO is a fairly specific term which I'd like to avoid including in our variable names unless we have to. If we use a more general component name it should be a small readability improvement and we can handle other possible types of details in the future that could be displayed in the same way. E.g. tenure XYZ attribute.

Maybe we could use this:

```
const `TenureDetail = ( { text } ) => {
  // ...
}
```
Then it would be called by parent components something like this:

```
<TenureDetail text=`TMO: ${ tmoName }` />

<TenureDetail text=`XYZ: ${ xyzAttribute }` />
```

NeilvB (approved):
Nicely done.

I'm keen for us to rename and slightly rework the `TMO` component because I think we'll benefit from a more generic name. But I'm happy for that to come in on a separate PR.

PR #221, opened 2021-06-10
Improve UI

EGiataganas:
### Description of change

Improve UI following Hackney Design System

* Delete govuk-frontend from package json
* Start using hackney's default button
* Update app and document to Hackney Design system
* Replace login/logout with signin/signout
* Update search component
* Typography:Body-text, replace govuk-body to lbh-body
* Typography:Table, add lbh-table on table tag
* Typography:Caption, add lbh-caption
* Typography:Headings, update to match to Hackney Design system
* Typography:Links, use/add lbh-link class
* Typography:Fonts, apply Hackney Design sytem bold
* Typography:List, apply Hackney Design sytem list
* Colours, Use Hackney Design system colours
* Use Table component
* Button, apply Hackney Design system secondary button

### Story Link

https://www.pivotaltracker.com/story/show/178395035


src/pages/_document.js
```diff
@@ -1,13 +1,18 @@
 import Document, { Html, Head, Main, NextScript } from 'next/document'
 
 class AppDocument extends Document {
+  static async getInitialProps(ctx) {
+    const initialProps = await Document.getInitialProps(ctx)
+    return { ...initialProps }
+  }
+
```
NeilvB, line 8: How come this was added? I couldn't see how it's related to styling changes..

NeilvB (approved):
The new UI is looking excellent.

Lots of careful, methodical work here.

My only question is about the spacing between the property information and the tabs that sit underneath it - see the tenure text and the tab in the image below. It looks like we may have lost spacing here somewhere. Is that intentional?

property-spacing

siapankina (approved):
looks great

src/pages/_document.js
EGiataganas: I've compared it from Hackney Design System [examples][1]

[1]:https://github.com/LBHackney-IT/hackney-design-system-examples/blame/main/next/pages/_document.js

EGiataganas:
Hackney Design system has a bit different spacing. For example `govuk` has `margin-bottom: 20px;` where hackney design system doesn't. In addition I've also wrapped all the links to `lbh-link` class which this has caused some spacing changing. I am in contact with Yasmeen and once we migrate fully some of the screens will changed.

PR #228, opened 2021-06-15
Show a work order list for DLO contractors

pixeltrix:
Hackney DLO contractors can raise repairs as well as carry them out so tweak the home page to return a list view when a user has both the agent role and the contractor role.


cypress/fixtures/work-orders/filter.json
```diff
@@ -0,0 +1,197 @@
+{
```
benbaumann95, line 1: This file should already exists under `fixtures/filter/work-order.json` but happy if we want to move the file location to this instead. Also I think the formatting looks slightly off just at a glance.

pixeltrix: Ugh, I just saw that the request wasn't being intercepted and assumed that it didn't exist - will update to point to existing file.

pixeltrix:
Updated to Cypress 7.5.0 and converted `cy.route` to `cy.intercept`

NeilvB (approved):

PR #232, opened 2021-06-15
Add prompt message to trade search

siapankina:
ticket : https://www.pivotaltracker.com/story/show/178484940

Screenshot 2021-06-15 at 15 30 30

### DIFF
--- a/src/components/Property/RaiseRepair/__snapshots__/RaiseRepairForm.test.js.snap
+++ b/src/components/Property/RaiseRepair/__snapshots__/RaiseRepairForm.test.js.snap
@@ -74,6 +74,10 @@ exports[`RaiseRepairForm component should render properly 1`] = `
             >
               *
             </span>
+             
+            <span>
+               - Search by type (e.g. Gas) or code (e.g. GS)
+            </span>
           </label>
           <input
             autocomplete="off"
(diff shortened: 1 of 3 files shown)


src/components/WorkElement/TradeDataList.js
```diff
@@ -10,6 +10,7 @@ const TradeDataList = ({ trades, register, errors, onTradeSelect }) => {
     <DataList
       name="trade"
       label="Trade"
+      labelMessage="- Search by type (e.g. Gas) or code (e.g.GS)"
```
benbaumann95, line 13: Tiny nitpick:
I know the screenshot has it like you have it here but should there not be a space between `e.g.` and `GS` like we have in `e.g. Gas` ?
Seems inconsistent otherwise

src/components/Property/RaiseRepair/__snapshots__/RaiseRepairForm.test.js.snap
```diff
@@ -74,6 +74,10 @@ exports[`RaiseRepairForm component should render properly 1`] = `
             >
               *
             </span>
+             
+            <span>
+               - Search by type (e.g.Gas) or code (e.g.GS)
```
benbaumann95, line 79: Ah so I meant it should be `e.g. Gas` and `e.g. GS` 😃

PR #236, opened 2021-06-15
Fetch work order when closing

NeilvB:
Otherwise it's possible to change the work order ID in the url and
show the close page for a non-existent work order. This surfaces
the issue earlier by showing an error message and will support
later work which requires additional props from the work order at the
point of closure.

Also:
  - Restructured a beforeEach in an integration test to avoid making
  network requests

Supports upcoming work for https://www.pivotaltracker.com/n/projects/2500129/stories/178315084

### DIFF
--- a/src/components/WorkOrders/CloseWorkOrder.js
+++ b/src/components/WorkOrders/CloseWorkOrder.js
@@ -1,6 +1,6 @@
 import PropTypes from 'prop-types'
 import CloseWorkOrderForm from './CloseWorkOrderForm'
-import { useState } from 'react'
+import { useState, useEffect } from 'react'
 import { convertToDateFormat } from '../../utils/date'
 import SummaryCloseWorkOrder from './SummaryCloseWorkOrder'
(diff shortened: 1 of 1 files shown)


benbaumann95 (approved):

src/components/WorkOrders/CloseWorkOrder.js
```diff
@@ -70,7 +101,7 @@ const CloseWorkOrder = ({ reference }) => {
         <>
           {CloseWorkOrderFormPage && (
             <CloseWorkOrderForm
-              reference={reference}
+              reference={workOrder?.reference}
```
benbaumann95, line 104: If we aren't using the `workOrder` object anywhere else probably could just set the reference directly from the `await getWorkOrder(reference)` call to avoid having to use a safe operator here but not significant at all

NeilvB: I've reworked this PR to avoid using this operator.

PR #242, opened 2021-06-17
Show link to DRS web booking manager for DRS orders (low priority) without an appointment

benbaumann95:
### Description of change

Upon creation of a DRS work order, you have a link to schedule an appointment via DRS web booking manager on the success page. However, if you navigate off this page, then there is no longer an ability to navigate to the relevant url. This PR adds this link on the work order view page, under the appointment details. A schedulerSessionId is created or used from the existing cookie to append to this url which is needed for access to DRS. 
The url is taken from the `workOrder/:id` endpoint and presents the link if it is there (for low priority work orders)

- Show link to DRS web booking manager for DRS orders (low priority) without an appointment
- Do not show link to schedule and appointment for immediate or emergency priority work orders
- Add AppointmentDetails snapshot tests with the relevant priority and appointment scenarios

### Story Link

https://www.pivotaltracker.com/n/projects/2500129/stories/178552315


EGiataganas (approved):
Great work!

NeilvB (approved):

PR #268, opened 2021-06-28
Show authorisation pending approval filter if user role includes authorisation manager

benbaumann95:
### Description of change

We used to hide the authorisation pending approval filter if the user had a contractors role associated with them. However, a DLO operative may have a DLO contractor role in addition to an authorisation manager role or agent role.

This change will display the filter if the user has a DLO role regardless of the other groups they may be in.

### Story Link

https://www.pivotaltracker.com/n/projects/2500129/stories/178690726


siapankina (approved):
😎

benbaumann95:
> 😎

Thanks for the review. Just waiting on Sarah to confirm something

EGiataganas (approved):

PR #277, opened 2021-07-01
Abstract away user permission structure

siapankina:


benbaumann95 (commented):
Ooh this is already looking much better than what we had, nice job!
Think you just have a commit "wip" that could probably get squashed/fixup and I think a recent change to the filter permissions means that we need to update `canSeeAllFilters` accordingly

benbaumann95 (approved):
🎉  Nice adding the tests too

PR #278, opened 2021-07-02
Improve performance

benbaumann95:
### Description of change

- Add lru-cache dependency
- Add caching middleware for service API requests
- For now just validate caching only for certain requests:
-- Cache data that is unlikely to change regularly
-- Do not cache data that will change frequently such as all the DRS and work order related information

### Story Link

https://www.pivotaltracker.com/n/projects/2500129/stories/178707947

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "http-status-codes": "^2.1.4",
     "jsonwebtoken": "^8.5.1",
     "lbh-frontend": "^3.5.3",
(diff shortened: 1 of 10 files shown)


siapankina (approved):
Ben, it looks great!
Maybe would be good for Evangelos or Neil have a look as well  ;)

EGiataganas (approved):

PR #285, opened 2021-07-09
Extract common use of Panel component to Template

EGiataganas:
### Description of change

Extract common use of Panel component to Template

#### Before
Screenshot 2021-07-05 at 11 55 53
Screenshot 2021-07-09 at 11 40 54

#### After
Screenshot 2021-07-09 at 12 12 48
Screenshot 2021-07-09 at 12 15 13


siapankina (approved):
well done! Really useful refactoring
thanks 👍

PR #286, opened 2021-07-15
Work order header enhancements 

siapankina:
Screenshot 2021-07-15 at 08 06 35
Screenshot 2021-07-15 at 09 44 04

Screenshot 2021-07-15 at 11 42 29


src/components/WorkOrder/WorkOrderHeader.js
```diff
@@ -27,6 +28,15 @@ const WorkOrderHeader = ({
     return currentTime > appointmentStartTime
   }
 
+  const completionReason = () => {
+    if (workOrder.status == 'Work Complete') {
+      return 'Completed'
+    }
+    if (workOrder.status == 'No Access') {
+      return 'No Access'
+    }
```
EGiataganas, line 37: How about using three equals signs `workOrder.status === REASON`.

src/components/WorkOrder/WorkOrderHeader.js
```diff
@@ -55,6 +65,16 @@ const WorkOrderHeader = ({
         <div className="lbh-body-xs">
           <span>Assigned to: {workOrder.owner}</span>
         </div>
+        {workOrder.closedDated && (
+          <div className="lbh-body-xs">
+            <span>
+              <strong>
+                {completionReason()}:{' '}
+                {formatDateTime(new Date(workOrder.closedDated))}
```
EGiataganas, line 73: How about combine these 2 lines together, like that:
```
{`${completionReason()}:  ${formatDateTime(new Date(workOrder.closedDated))}`}
```

src/components/WorkOrder/WorkOrderInfo.js
```diff
@@ -1,18 +1,30 @@
...
+  EMERGENCY_PRIORITY_CODE,
+} from '../../utils/helpers/priorities'
 import { formatDateTime } from '../../utils/time'
 
 const WorkOrderInfo = ({ workOrder }) => {
+  const highPriority = () => {
+    if (
+      workOrder.priorityCode == EMERGENCY_PRIORITY_CODE ||
+      workOrder.priorityCode == IMMEDIATE_PRIORITY_CODE
+    ) {
```
EGiataganas, line 13: Keep the three equals signs and how about doing it like that:
```
const highPriority = () => {
  return (
    workOrder.priorityCode == EMERGENCY_PRIORITY_CODE ||
    workOrder.priorityCode == IMMEDIATE_PRIORITY_CODE
  )
}
```

EGiataganas:
Well done, this is a great job. I left some comment to think about 😄

EGiataganas (approved):

src/components/WorkOrder/WorkOrderHeader.js
```diff
@@ -27,6 +28,15 @@ const WorkOrderHeader = ({
     return currentTime > appointmentStartTime
   }
 
+  const completionReason = () => {
```
NeilvB, line 31: A small improvement here would be to just do the check for the special case of `workOrder.status === 'Work Complete'` and otherwise just return whatever is in `workOrder.status`. Then if more completion statuses are added to the system later this we don't need to worry about updating this function.

src/components/WorkOrder/WorkOrderHeader.js
```diff
@@ -55,7 +65,16 @@ const WorkOrderHeader = ({
         <div className="lbh-body-xs">
           <span>Assigned to: {workOrder.owner}</span>
         </div>
-
+        {workOrder.closedDated && (
```
NeilvB, line 68: We would benefit from a test (snapshot test probably) for this behaviour.

src/styles/components/circle-tag.scss
```diff
@@ -0,0 +1,40 @@
...
+  }
+  &--status-pending-approval {
+    background-color: repairs-hub-colour('yellow');
+  }
+  &--status-variation-pending-approval {
+    background-color: repairs-hub-colour('yellow');
+  }
+  &--status-authorisation-pending-approval {
+    background-color: repairs-hub-colour('yellow');
+  }
```
NeilvB, line 30: Very small nitpick but we could reduce the above three rules into one rule with three selectors if we wanted..

src/components/WorkOrder/WorkOrderHeader.js
```diff
@@ -29,12 +29,7 @@ const WorkOrderHeader = ({
   }
 
   const completionReason = () => {
-    if (workOrder.status == 'Work Complete') {
-      return 'Completed'
-    }
-    if (workOrder.status == 'No Access') {
-      return 'No Access'
-    }
+    workOrder.status === 'Work Complete' ? 'Completed' : workOrder.status
```
NeilvB, line 32: We need a `return` at the start of this line please, otherwise we'll be getting `undefined` back from the function

src/components/WorkOrder/__snapshots__/WorkOrderHeader.test.js.snap
```diff
@@ -0,0 +1,493 @@
...
+        <div
+          class="lbh-body-s govuk-!-margin-bottom-2"
+        >
+          <span
+            class="circle-tag circle-tag--status-work-complete"
+          />
+          <span
+            class="lbh-!-font-weight-bold"
+          >
+            Status: Work Complete
```
NeilvB, line 246: I think this will need to be `Status: Complete`. This might be resolved by the addition of the `return` to the helper function mentioned above.

src/components/WorkOrder/WorkOrderHeader.test.js
```diff
@@ -0,0 +1,130 @@
+import { render } from '@testing-library/react'
+import WorkOrderHeader from './WorkOrderHeader'
+import UserContext from '../UserContext/UserContext'
+import { agent } from 'factories/agent'
+
+describe('WorkOrderHeader component', () => {
```
NeilvB, line 6: Thanks for increasing our test coverage! 👍

src/components/WorkOrder/WorkOrderHeader.js
siapankina: 🤦‍♀️

src/components/WorkOrder/WorkOrderHeader.js
siapankina: {''} was added automatically with prettier:fix

src/components/WorkOrder/__snapshots__/WorkOrderHeader.test.js.snap
NeilvB: On revisiting this I realised I wasn't looking in the right place in the snapshot for the output of that function. I can see that we show the expected text [here](https://github.com/LBHackney-IT/repairs-hub-frontend/pull/286/files#diff-6de5dff3be570d9c44b4d49d8ba242529217710cbb3cc6f88324e22866f28f11R322).

NeilvB (approved):

PR #288, opened 2021-07-15
Reschedule appointment

NeilvB:
Permit rescheduling in RH directly or through DRS.

[Pivotal Card](https://www.pivotaltracker.com/n/projects/2500129/stories/178480897)

### UI for RH Rescheduling

image

User then books appointment as per current flow. Then..

image

### Note added when opening Web Booking Manager

image


EGiataganas (approved):

siapankina (approved):
great job 😎

PR #289, opened 2021-07-15
Ensure correct work order reference formatting

Asura5:
The historical work orders from UH have integer primary keys e.g. "246282" however for searching and display in Repairs Hub, 8 the references have a confirmed length of eight characters (e.g. 00246282) - this change maintains the consistency between the search term, the route, and the display


src/components/WorkOrder/WorkOrderDetails.js
```diff
@@ -41,7 +41,7 @@ const WorkOrderDetails = ({
       <GridRow>
         <GridColumn width="two-thirds">
           <h1 className="lbh-heading-h1 display-inline govuk-!-margin-right-6">
-            Work order: {workOrder.reference}
+            Work order: {workOrder.reference.toString().padStart(8, '0')}
```
siapankina, line 44: would you mind adding a test for that to WorkOrderDetails.test.js for this behaviour? (to have a snapshot of it)

src/components/WorkOrder/WorkOrderDetails.test.js
```diff
@@ -102,6 +119,24 @@ describe('WorkOrderDetails component', () => {
       )
       expect(asFragment()).toMatchSnapshot()
     })
+    it('should render migrated work order references correctly', () => {
```
siapankina, line 122: maybe we could add a comment here so it is clear how is "correctly": 
// if work order is 648707 then it should render is as 0064 8707 

something like that

siapankina (approved):
Tiny nitpick ;) Otherwise looks great

src/components/WorkOrder/WorkOrderDetails.js
siapankina: Evangelos has a good point, were you able to discuss that possibility with b/e developers?

src/components/WorkOrder/WorkOrderDetails.test.js
Asura5: Added

src/components/WorkOrder/WorkOrderDetails.js
Asura5: This is a good idea, and worth doing - though throughout the backend, the work order reference is an int - so it would need a substantial refactor. Therefore worth keeping this change too in my opinion

PR #290, opened 2021-07-16
Don't show appointment link if w/o is closed

siapankina:
when work orders is closed user should not  see " schedule appointment link 

ticket: https://www.pivotaltracker.com/story/show/178895770


src/components/WorkOrder/AppointmentDetails.js
```diff
@@ -16,6 +18,16 @@ import {
 const AppointmentDetails = ({ workOrder, schedulerSessionId }) => {
   const { user } = useContext(UserContext)
 
+  const statusAllowsScheduling = (currentStatus) => {
+    const statuses = [
+      STATUS_CANCELLED.description,
+      STATUS_AUTHORISATION_PENDING_APPROVAL.description,
+      STATUS_COMPLETE.description,
+      STATUS_NO_ACCESS.description,
+    ]
```
NeilvB, line 27: I like this separate function.

This is a minor suggestion and doesn't block a merge but another way to write this particular part of it would be to map over the statuses and return each description. 

The benefit is that the reader can just read one line to see that we're getting the description for every single status. At the moment someone could add something like `SOME_NEW_STATUS.code` in this array when they mean `.description` and cause a tricky-to-find bug. Using the map approach makes it clearer we're just always interested in the `description` and expresses that in just one place.

E.g.

```
  // ...

  const closedStatusDescriptions = [
    STATUS_CANCELLED,
    STATUS_AUTHORISATION_PENDING_APPROVAL,
    STATUS_COMPLETE,
    STATUS_NO_ACCESS
  ].map((status) => status.description })

  return !closedStatusDescriptions.includes(currentStatus)

  // ...
```

NeilvB (changes requested):
This is looking good so far but we should add a couple of snapshot tests please.

I don't think it's worthwhile going through every possible closed state but to have one for a non-completed status and one for a completed status would be good reassurance.

E.g. something like this

```
      describe('when the work order is not closed', () => {
        it('shows a schedule link', () => { //... })
      })

      describe('when the work order is in a closed state', () => {
        it('does not show a schedule link', () => { // ... })
      })
```

src/components/WorkOrder/AppointmentDetails.js
EGiataganas: I agree with @NeilvB! I am wondering if this best to moved to `src/utils/status-codes.js`, so we have all the related functions in one place.

src/utils/status-codes.js
```diff
@@ -33,6 +33,11 @@ export const STATUS_VARIATION_REJECTED = {
   description: 'Variation Rejected',
 }
 
+export const STATUS_NO_ACCESS = {
+  code: 1000,
+  description: 'No Access',
+}
+
```
EGiataganas, line 40: You will also need to include `STATUS_NO_ACCESS.description`, at the end of this file.

NeilvB (approved):

EGiataganas (approved):

PR #291, opened 2021-07-16
Show appointments for any priority work order

NeilvB:
[Pivotal Tracker card](https://www.pivotaltracker.com/n/projects/2500129/stories/178724846)

Screenshot showing an immediate priority work order with an appointment.

image


siapankina (approved):
🚢

PR #293, opened 2021-07-19
Fix snapshot test setups

NeilvB:
Looks like there may have been scoping issues previously, causing
test setups and expectations not to match descriptions. This reworks
test setups to make it a bit clearer what properties of a work order are
being set for each test.

Also updates a couple of snapshots which did not reflect the relevant
test descriptions, but now do following the test setup changes.


EGiataganas (approved):

siapankina (approved):
:shipit:

PR #295, opened 2021-07-19
Reduce duplications across functions

siapankina:


NeilvB:
There's more simple util functions which would benefit from this treatment.

Please could we do the same for the simplest util functions like `getPriorities`, `getTrades`, `getSchedulerSessionId`, `getNotes`, `getTasksAndSors` and others.

We wouldn't need to add any more tests to `requests.test.js` for these.

NeilvB:
There's a few more lurking.. e.g. `getVariationTasks`,`postJobStatusUpdate`, the ones for properties and several others. Would be good to get all the simple ones in one PR.

NeilvB (approved):

PR #296, opened 2021-07-20
Work order model

NeilvB:
### Description of change

Introduce a [simple work order model](https://github.com/LBHackney-IT/repairs-hub-frontend/pull/296/files#diff-bdffb02560ef35f250b69659cbfabd689e2f7c8c82c0bacdf784f528c100f3e4) with various getter-type methods.

Rework some unit tests so there's less need for them to be concerned with specific statuses and priorities, and instead stub out the new getter methods.

### DIFF
--- a/cypress/integration/raise-repair/appointment-form.spec.js
+++ b/cypress/integration/raise-repair/appointment-form.spec.js
@@ -281,6 +281,9 @@ describe('Schedule appointment form', () => {
         .contains('Book appointment')
         .click({ force: true })
       //appointment api check
+
+      cy.wait('@apiCheckAppointment')
+
       cy.get('@apiCheckAppointment')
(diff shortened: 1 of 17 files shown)


cypress/integration/raise-repair/appointment-form.spec.js
```diff
@@ -281,6 +281,9 @@ describe('Schedule appointment form', () => {
         .contains('Book appointment')
         .click({ force: true })
       //appointment api check
+
+      cy.wait('@apiCheckAppointment')
```
siapankina, line 285: all hail wait 🤣

siapankina (approved):
Noice 😎

EGiataganas (approved):

PR #299, opened 2021-07-21
Hide MultiButton if work order is closed

siapankina:


src/components/WorkOrder/WorkOrderDetails.test.js
```diff
@@ -102,6 +102,30 @@ describe('WorkOrderDetails component', () => {
       )
       expect(asFragment()).toMatchSnapshot()
     })
+    it('should not render Multibutton if work order is closed', () => {
```
NeilvB, line 105: This is a bit of nitpick but it's a best practice IMO to avoid referring only to generic names for component or other generic units of code in test descriptions. It's unlikely I admit, but we could in the future add an additional multibutton, for some other functionality, which is always visible. That would then make this test description incorrect and might confuse the reader. Alternatively, we might replace the existing multibutton in the future with another implementation like a set of separate action buttons and this test description could lead to confusion unless we remember to update it.

I suggest we call it something more specific to repairs hub like 'should not render the work order action menu ...'

src/components/WorkOrder/WorkOrderDetails.test.js
```diff
@@ -102,6 +102,30 @@ describe('WorkOrderDetails component', () => {
...
+      const { asFragment } = render(
+        <UserContext.Provider value={{ user: agent }}>
+          <WorkOrderDetails
+            propertyReference={props.property.propertyReference}
+            workOrder={
+              new WorkOrder({
+                ...workOrderData,
+                status: 'No Access',
```
NeilvB, line 113: The four new unit tests in this file only seem to differ by their `status` - is that correct? If so I can't see how that would change the behaviour of the component under test. It seems to me that we can safely remove all but one of the tests, remove the status, and just set `closedDated`.

NeilvB (changes requested):
Requesting we take another look at the tests as there may be some duplication.

src/components/WorkOrder/WorkOrderDetails.test.js
NeilvB: Or 'should not render the work order action multibutton...' would also be fine. It's just to avoid referencing just the generic component on its own. Nitpick essay over!

NeilvB (approved):

EGiataganas (approved):

PR #300, opened 2021-07-21
Warning when scheduling on appointment day

NeilvB:
Part two of [story](https://www.pivotaltracker.com/n/projects/2500129/stories/178480897)

Specifically regarding adding a warning message for when today is the appointment today (see [comment](https://www.pivotaltracker.com/story/show/178480897/comments/225606215))

#### UI

image

### DIFF
--- a/src/components/WorkOrder/AppointmentDetails.js
+++ b/src/components/WorkOrder/AppointmentDetails.js
@@ -1,7 +1,6 @@
 import { useContext } from 'react'
 import PropTypes from 'prop-types'
 import UserContext from '../UserContext/UserContext'
-import Link from 'next/link'
 import { dateToStr } from '../../utils/date'
 import { STATUS_CANCELLED } from '../../utils/status-codes'
 import {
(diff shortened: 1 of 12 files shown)


src/components/WorkOrder/AppointmentDetails.js
```diff
@@ -41,36 +42,24 @@ const AppointmentDetails = ({ workOrder, schedulerSessionId }) => {
...
       return (
-        <Link href={href}>
-          <a className="lbh-link">{linkText}</a>
-        </Link>
+        <ScheduleInternalAppointmentLink
+          workOrderReference={workOrder.reference}
+          hasExistingAppointment={hasExistingAppointment}
+          appointmentIsToday={workOrder.appointmentIsToday()}
+        />
       )
```
EGiataganas, line 63: That's a great improvement moving them into separate components. I am wondering if it's worth to add another component `ScheduleAppointmentLink` which will have the if logic, so we have only one component here.

NeilvB: Thanks for this - I think this is a great shout but are you OK if I implement this in a separate PR another time? 

I'm keen to get this story over to QA as it has been delayed for a few different reasons.

siapankina (approved):
🚢

src/components/WorkOrder/AppointmentDetails.js
siapankina: Agree with Evangelos, and it is fine by me if we implement it in a separate PR

EGiataganas: Yeah sure, we can pick up this on a separate PR.

EGiataganas (approved):

EGiataganas (approved):

EGiataganas (approved):

PR #315, opened 2021-08-02
[WIP]Implement automatic percentage calculation

siapankina:
in case operatives were assigned:
  - 1 operative, shows 100% as a default and can't be changed
  - 2 operatives, shows 50%/50%
  - 3 operatives, shows 33.33% each

when no operatives assigned and selected 1 operative it automatically shows 100% as default
when more than 1, it shows '-'

remove button that allows removing assigned operatives

### DIFF
--- a/src/components/Operatives/SelectOperatives.js
+++ b/src/components/Operatives/SelectOperatives.js
@@ -1,5 +1,6 @@
 import PropTypes from 'prop-types'
 import { useEffect, useState } from 'react'
+import cx from 'classnames'
 import OperativeDataList from './OperativeDataList'
 import SelectPercentage from '../WorkOrders/SelectPercentage'
 import { GridRow, GridColumn } from '../Layout/Grid'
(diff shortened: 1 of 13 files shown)


src/components/Operatives/SelectOperatives.js
```diff
@@ -1,60 +1,179 @@
...
+  const calculateTotalPercentage = (
+    selOperatives,
+    operativesIndexPercentages
+  ) => {
+    //not using rounding logic in case options are (33.3, 33.3 and 30, then it will round it to 100, and it is not correct)
+    if (
+      selOperatives.length === 3 &&
+      operativesIndexPercentages[0] === '33.3%' &&
+      operativesIndexPercentages[1] === '33.3%' &&
+      operativesIndexPercentages[2] === '33.3%'
```
NeilvB, line 59: We can shorten the above three lines to `operativesIndexPercentages.every(p => p === '33.3%')`

src/components/WorkOrders/CloseWorkOrder.js
```diff
@@ -141,19 +147,36 @@ const CloseWorkOrder = ({ reference }) => {
     setCloseWorkOrderFormPage(!CloseWorkOrderFormPage)
   }
 
+  const updateTotalPercentage = (percentageNumber) => {
+    setTotalPercentage(percentageNumber)
+  }
```
NeilvB, line 152: I'm not sure if we gain much from wrapping `setTotalPercentage` here in another function. Could we just pass `setTotalPercentage` down to the children (and name the props on them to match)? Keeping the convention makes sense to me unless we have a reason not to.

siapankina:
in favor https://github.com/LBHackney-IT/repairs-hub-frontend/pull/325

PR #321, opened 2021-08-20
Refactor operative selection & align with react hook form approach

NeilvB:
React Hook Form encourages the use of uncontrolled components. The
previous implementation was an odd mix of partially controlled and
uncontrolled with state both in components and in the DOM.

- Remove state management within OperativeDataList and rely on the
  form parent to extract operative ids instead of using hidden fields.
- Use a defaultValue prop instead of value [1]

This should hopefully help us to rationalise this form and help us
when implementing percentages.

1 - see https://reactjs.org/docs/uncontrolled-components.html#default-values)

We now avoid the React warning about a component changing from controlled to uncontrolled

image

### DIFF
--- a/src/components/WorkOrders/CloseWorkOrder.js
+++ b/src/components/WorkOrders/CloseWorkOrder.js
@@ -27,6 +27,8 @@ const CloseWorkOrder = ({ reference }) => {
   const [CloseWorkOrderFormPage, setCloseWorkOrderFormPage] = useState(true)
   const router = useRouter()
 
+  const OPERATIVE_ID_REGEX = /\[(\d+)\]$/
+
   const makePostRequest = async (
     workOrderCompleteFormData,
     operativeAssignmentFormData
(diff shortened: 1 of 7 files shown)


src/components/WorkOrders/CloseWorkOrder.js
```diff
@@ -145,9 +147,15 @@ const CloseWorkOrder = ({ reference }) => {
...
-      .map(([, value]) => Number.parseInt(value))
+    const operativeIds = Object.keys(e)
+      .filter((k) => k.match(/operative-\d+/))
+      .map((operativeKey) => {
+        const matches = e[operativeKey].match(OPERATIVE_ID_REGEX)
+
+        if (Array.isArray(matches)) {
+          return Number.parseInt(matches[matches.length - 1])
+        }
+      })
```
siapankina, line 158: 👍

siapankina (approved):

PR #328, opened 2021-09-03
Show smv on work order page

siapankina:
*ticket: https://www.pivotaltracker.com/story/show/178775488

*screenshots :
*No changes in tasks and SORs
Screenshot 2021-09-03 at 12 18 52

* Tasks and SORs were updated, original table is hidden by default 
Screenshot 2021-09-03 at 12 20 28

*Original table is visible 
Screenshot 2021-09-03 at 12 20 45


src/components/Layout/Collapsible/Collapsible.js
```diff
@@ -1,14 +1,21 @@
 import PropTypes from 'prop-types'
-import { useState } from 'react'
+import { useState, useEffect } from 'react'
 
 const Collapsible = ({
   heading,
   children,
   contentMargin,
   collapsableDivClassName,
+  onTasksAndSorsTab,
```
NeilvB, line 9: This is functionally fine but it's a good practice for us to keep more generic components from knowing about our business logic as much as possible. In this case, knowing about what tab it's on via its prop name. All of the other props are really general names here and we should try to keep it like that :)

The underlying functionality we want here is for the parent to tell the collapsible whether to start open or closed. Any parent which doesn't specify this should get the existing open-by-default behaviour.

We can use a `startsClosed` prop with a more generic name and supply that as the initial state instead of using an effect hook to clean things up a little. If it's not passed in from the parent, the collapsible component will have the default open behaviour.

Please check out the `App.js` and `Collapsible.js` files in the sandbox here https://codesandbox.io/s/reverent-hermann-5gmw3?file=/src/Collapsible.js

NeilvB (changes requested):
Requesting that we extract a few util functions and add more tests to them, as well as rework the `Collapsible` element a bit too.

src/components/WorkOrder/TasksAndSors/TasksAndSorsView.js
```diff
@@ -9,23 +10,21 @@ const TasksAndSorsView = ({ tabName, tasksAndSors }) => {
     return !tasks.every(originalQuantityIsEqualToCurrentQuantity)
   }
```
NeilvB, line 11: Probably also worth extracting this one (maybe into a `tasks.js` util file) and getting some unit tests around it

NeilvB (changes requested):

NeilvB (approved):

PR #329, opened 2021-09-03
Add operative

EGiataganas:
### Description of change

Apply same pattern everywhere
-Use User helper methods
-Follow same pattern methods for User
-Use factories on Header test

Add operative role

Update user factories to have hasOperativePermissions to false

Allow operative role only log out and index page permission

Extract functionality to helper userPermission function

### Story Link

https://www.pivotaltracker.com/story/show/179177031


NeilvB (changes requested):
Very minor test copy request. Otherwise looks great.

NeilvB (approved):

PR #335, opened 2021-09-07
Operatives index page

EGiataganas:
### Description of change

- Operative's work orders view
- Export to OperativeWorkOrderListItem
- Rename schedule-warning class to warning-info-box to be more generic
- Extract WarningInfoBox to its own component

### Story Link

https://www.pivotaltracker.com/story/show/179177031


cypress/integration/home-page.spec.js
```diff
@@ -438,4 +438,73 @@ describe('Home page', () => {
...
+
+        cy.visit('/')
+        cy.wait('@operativesWorkOrders')
+      })
+
+      it('Displays work order appointments slots and priority', () => {
+        cy.get('.lbh-heading-h1').contains('Friday, 11 June')
+
+        cy.get('.lbh-list').within(() => {
+          cy.checkForAppointmentsDetails('08:00-13:00', '5 [N] NORMAL')
```
NeilvB, line 493: We're not getting much benefit from an extracted function here. It's just a couple of content checks.

I'd prefer to have this all contained within the test and no separate function please.

I don't generally worry about duplicate code in test expectations. It's more valuable to see explicitly within each test what it is checking without having to jump into separate files. I know we have functions for user sign in, but those are more for making the setup easier.

src/styles/components/arrow.scss
```diff
@@ -0,0 +1,26 @@
...
+
+.up {
+  transform: rotate(-135deg);
+  -webkit-transform: rotate(-135deg);
+}
+
+.down {
+  transform: rotate(45deg);
+  -webkit-transform: rotate(45deg);
+}
```
NeilvB, line 26: Please remove the unused CSS rules from this file.

src/utils/userPermissions.test.js
```diff
@@ -174,3 +175,27 @@ describe('canSeeWorkOrders', () => {
     })
   })
 })
+
+describe('canSeeOperativeWorkOrders', () => {
+  describe('when user allowed to access operarive work orders', () => {
```
NeilvB, line 180: Nitpick: typo -> `operative`

src/utils/userPermissions.test.js
```diff
@@ -174,3 +175,27 @@ describe('canSeeWorkOrders', () => {
...
+      name: 'Test Testerston',
+      email: 'test@test.com',
+      hasOperativePermissions: true,
+    }
+    it('returns true', () => {
+      expect(canSeeOperativeWorkOrders(user)).toBe(true)
+    })
+  })
+
+  describe('when user NOT allowed to access operarive work orders', () => {
```
NeilvB, line 191: Nitpick: typo -> `operative`

src/components/Operatives/OperativeWorkOrderListItem.js
```diff
@@ -0,0 +1,42 @@
+import PropTypes from 'prop-types'
+
+const OperativeWorkOrderListItem = ({ operativeWorkOrders }) => (
+  <>
+    <ol className="lbh-list">
+      {operativeWorkOrders.map((operativeWorkOrder, index) => (
```
NeilvB, line 6: It's more conventional to do the `.map` call in a parent component and supply an individual object as a `prop` to an "individual item" component. We should move the `<ol>` and `map` up into the parent please as it will make it easier to read.

See eg https://github.com/LBHackney-IT/repairs-hub-frontend/blob/develop/src/components/Properties/PropertiesTable.js#L22

NeilvB (changes requested):
Requesting we move the `map` into the parent component and remove some unused CSS

NeilvB (approved):
Looking good

PR #346, opened 2021-09-09
Operative work order page

siapankina:
***Story 
https://www.pivotaltracker.com/story/show/179282568

***screenshot
Screenshot 2021-09-09 at 21 52 14


cypress/integration/home-page.spec.js
```diff
@@ -379,13 +379,32 @@ describe('Home page', () => {
...
+              cy.contains('2 [E] EMERGENCY')
+            })
+            cy.get('.middle-right').within(() => {
+              cy.get('.arrow').should(
+                'have.attr',
+                'href',
+                '/work-orders/10000625'
+              )
+            })
           })
```
NeilvB, line 408: Please add a step to this test which clicks on a work order, then views the page to give us confidence we're rending our new set of elements / components instead of the original ones.

Will of course need request stubbing to populate the page.

I suggest just expecting some basic content including some simple checks for things which should appear for operatives and not for other users.

This should be okay

```
cy.contains('WO 10000625')

cy.get('div[class*="Multibutton"]').should('not.exist')
```

NeilvB (changes requested):

src/components/Operatives/OperativeWorkOrderListItem.js
```diff
@@ -1,24 +1,30 @@
...
-    </li>
-  </>
-)
+          <p className="lbh-body lbh-!-font-weight-bold text-white">
+            {operativeWorkOrder.priority}
+          </p>
+        </div>
+        <div className="middle-right">
+          <Link href={`/work-orders/${operativeWorkOrder.reference}`}>
+            <a className="arrow right" id="arrow-right"></a>
```
NeilvB, line 17: We should avoid adding an `id` as this element is in a list and ids should appear in the document only once.

NeilvB (approved):

PR #349, opened 2021-09-12
Attend work order

NeilvB:
### Description of change

Allow operatives to attend (i.e. set to complete or no access) a work order.


siapankina (approved):
Looks great!

PR #363, opened 2021-09-15
Styling improvements

EGiataganas:
Current                    |  Proposed
:-------------------------:|:-------------------------:
cur1 | prop1
cur2 | prop2


src/styles/typography.scss
```diff
@@ -1,5 +1,28 @@
...
+    line-height: 0.875rem;
+    margin-top: 1em;
+  }
+}
+
+.lbh-button__start-icon {
+  @include govuk-media-query($until: tablet) {
+    display: none;
+  }
+}
```
NeilvB, line 28: Think it's best to keep this file just for any general font-specific styling we want like font size and weight which applies broadly, otherwise it's a bit mixed with spacing and display rules.

NeilvB (changes requested):
Small request to move some rules out of `typography.scss` and then good to go

NeilvB (approved):

PR #384, opened 2021-09-30
Collapse description text if more than 3 lines

siapankina:
ticket: https://www.pivotaltracker.com/story/show/179684602
Screenshot 2021-09-30 at 13 23 51

Screenshot 2021-09-30 at 13 25 07
Screenshot 2021-09-30 at 13 25 21

### DIFF
--- a/src/components/Operatives/OperativeWorkOrderDetails.js
+++ b/src/components/Operatives/OperativeWorkOrderDetails.js
@@ -4,9 +4,12 @@ import { longMonthWeekday } from '../../utils/date'
 import { WorkOrder } from '../../models/workOrder'
 import { GridColumn, GridRow } from '../Layout/Grid'
 import { useRouter } from 'next/router'
+import { useCallback, useState } from 'react'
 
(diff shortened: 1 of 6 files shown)


src/components/Operatives/OperativeWorkOrderDetails.js
```diff
@@ -57,8 +76,35 @@ const OperativeWorkOrderDetails = ({ property, workOrder, personAlerts }) => {
           </GridColumn>
         </GridRow>
         <div className="lbh-heading-h5">Description</div>
-        <p className="govuk-body">{workOrder.description}</p>
-
+        <p
+          ref={pRef}
+          className={`govuk-body govuk-!-margin-bottom-0 ${
```
EGiataganas, line 81: Do we need the addition of `govuk-!-margin-bottom-0` class?

siapankina: if we don't want to have big gap between text and "show more/less", then yes

EGiataganas (approved):

PR #386, opened 2021-10-04
develop --> main

siapankina:


NeilvB (approved):

src/components/Operatives/OperativeWorkOrderDetails.js
```diff
@@ -146,6 +198,7 @@ OperativeWorkOrderDetails.propTypes = {
   property: PropTypes.object.isRequired,
   workOrder: PropTypes.instanceOf(WorkOrder).isRequired,
   personAlerts: PropTypes.array.isRequired,
+  locationAlerts: PropTypes.array.isRequired,
 }
 
 export default OperativeWorkOrderDetails
```
EGiataganas, line 204: Since we are using in both places `getCautionaryAlertsType(locationAlerts, personAlerts)`, we can create a variable and store it. Something like that:
```js
const cautionaryAlertsType = getCautionaryAlertsType(locationAlerts, personAlerts)
```

By doing that on Line 121, it can be
```js
cautionaryAlertsType && ( etc )
```

siapankina: what this storing will give us?

siapankina: ok, I see

EGiataganas: We won’t call twice the `getCautionaryAlertsType(locationAlerts, personAlerts)`

PR #388, opened 2021-10-05
Update cautionary alerts list

siapankina:
- Updated cautionary alerts list : added missed location alerts
- Implemented the logic that Evangelos suggested --> https://github.com/LBHackney-IT/repairs-hub-frontend/pull/385/files#diff-f63084e601cef55a9fdcb9cb76eb15e3f39a47741118835a665ba6ce3f40a65dR13-R23
- Refactored code to keep it DRY and reusable 
- Added test


src/components/CautionaryContact/CautionaryContactTable.test.js
```diff
@@ -2,7 +2,7 @@ import { render } from '@testing-library/react'
 import UserContext from '../UserContext'
 import CautionaryContactTable from './CautionaryContactTable'
 import { operative } from 'factories/operative'
-import { CAUTIONARY_CONTACT } from '../../utils/cautContactAlertsDescription'
+import { CAUTIONARY_CONTACT } from '../../utils/cautContactAlerts'
```
NeilvB, line 5: I know this wasn't named in this PR but would prefer this to be `CAUTIONARY_CONTACTS` to indicate we're working with an array.

src/components/CautionaryContact/CautionaryContactView.js
```diff
@@ -1,6 +1,6 @@
 import BackButton from '../Layout/BackButton'
 import CautionaryContactTable from './CautionaryContactTable'
-import { CAUTIONARY_CONTACT } from '../../utils/cautContactAlertsDescription'
+import { CAUTIONARY_CONTACT } from '../../utils/cautContactAlerts'
 
 const CautionaryContactView = (searchQuery) => {
   const cautionaryContacts = CAUTIONARY_CONTACT
```
NeilvB, line 6: We could just pass the `CAUTIONARY_CONTACTS` straight in to the component rather than define a new const for a slight readability improvement.

src/utils/cautContactAlerts.test.js
```diff
@@ -0,0 +1,32 @@
+import { getCautionaryAlertsType } from './cautContactAlerts'
+
+describe('getCautionaryAlertsType', () => {
+  it('combines two arrays of alerts and returns types of alerts', () => {
```
NeilvB, line 4: Nitpick but would be good to include the fact that we expect the output to remove duplicate types across the two arrays in this test description - that's an important part of the functionality.

NeilvB (changes requested):
Requesting the exported array const rename, otherwise looking good.

src/components/CautionaryContact/CautionaryContactView.js
EGiataganas: I would recommend to use the name of `CAUTIONARY_ALERTS` as the `CAUTIONARY_CONTACTS` indicated only the `personal_alerts` but we need to include `location_alerts` as well

siapankina: Thank you both ;)

src/components/CautionaryContact/CautionaryContactTable.test.js
EGiataganas: Yes I agree, we've discussed this again but due to the fact in designs were using singular @siapankina suggested to leave it the same. I think we should use plural and suggest to design team to update designs and have it everywhere plural. After discussing with Yasmeen we agreed on using `CAUTIONARY_ALERTS` due to the fact will include personal and location alerts

NeilvB (approved):

src/utils/cautContactAlerts.js
```diff
@@ -95,4 +95,32 @@ export const CAUTIONARY_CONTACT = [
...
+export const getCautionaryAlertsType = (locationAlerts, personAlerts) => {
+  let cautionaryAlerts = [...locationAlerts, ...personAlerts].map(
+    (cautionaryAlert) => cautionaryAlert.type
+  )
+  let uniqueCautionaryAlertsType = [...new Set(cautionaryAlerts)]
+
+  return uniqueCautionaryAlertsType.length > 1
+    ? uniqueCautionaryAlertsType.join(', ')
+    : uniqueCautionaryAlertsType.join('')
+}
```
EGiataganas, line 126: How about renaming the file to `cautionaryAlerts`, as it will include both `personal` and `location`?

PR #399, opened 2021-10-18
Extract the logic to a separate component OperativeWorkOrderView

EGiataganas:
### Description of change

Extract the logic to a separate component OperativeWorkOrderView


NeilvB (commented):
Is this breaking our convention for how we've been using `...View.js` files in the application?

As it stands, this PR would mean that we have one `...View.js` actually calling a child `...View.js` in certain conditions, which I don't think fits how we've used them up until now. I think they have been used as a single layer of top level components called by our pages.

If we want to add a new view file, I think a better approach would be to make both of them handle the data fetching and errors by extracting the main shared logic (i.e. `getWorkOrderView`) into a shared place used by both views, and then have the page component (i.e. `WorkOrderPage`) look at the current user's role to conditionally display the right page for them, instead of having that condition in the view.

PR #400, opened 2021-10-18
Update SMVs

siapankina:


src/components/WorkOrders/CloseWorkOrderForm.test.js
```diff
@@ -25,6 +25,7 @@ describe('CloseWorkOrderForm component', () => {
     availableOperatives: operatives,
     reason: 'No Access',
     selectedPercentagesToShowOnEdit: [],
+    totalSMVs: 76,
```
NeilvB, line 28: nitpick: typo -> `...SMV` and below

src/utils/helpers/calculations.test.js
```diff
@@ -87,3 +89,12 @@ describe('calculateTotalVariedCost', () => {
     expect(calculateTotalVariedCost(arrayOfObjects)).toEqual(300)
   })
 })
+
+describe('calculateSMV and strippingZeroes', () => {
+  it('returns 25.31', () => {
```
NeilvB, line 94: Test descriptions should include a description of how we expect the unit of code will decide what to output. It makes it easier for someone reading the test log output as to work out what might be going wrong in the event of a failure.

So this should be something like 'returns a number matching the percentage proportion including decimal points', and the next one should be something like 'returns a number matching the percentage proportion without decimals when none are calculated'. Bit of a mouthful but I think it gets the idea across a bit more clearly and should help a future reader.

PR #403, opened 2021-10-20
Allow task variation by operatives (part I)

neilvanbeinum:
### Description of change

Allow operatives to update tasks on their work order.

- [ ] Adding a comment is mandatory before clicking confirm to move to next step - NOT INCLUDED IN THIS PR
- [x] Can click on each SOR code and be taked to the update SOR code page
- [x] Can modify the quantity to increase of the selected SOR code
- [x] Can click confirm and be taken back to the previous page and see the updated quantity in the table
- [x] The total SMV is updated

### Story Link

https://www.pivotaltracker.com/n/projects/2500129/stories/179318892

### Screenshots

image

image

### DIFF
--- a/src/pages/work-orders/[id]/tasks/[taskId]/edit.js
+++ b/src/pages/work-orders/[id]/tasks/[taskId]/edit.js
@@ -1,14 +1,14 @@
 import { OPERATIVE_ROLE } from '../../../../../utils/user'
 import Meta from '../../../../../components/Meta'
 import { getQueryProps } from '../../../../../utils/helpers/serverSideProps'
-import UpdateTaskView from '../../../../../components/WorkOrder/TasksAndSors/UpdateTaskView'
+import EditTaskForm from '../../../../../components/WorkOrder/TasksAndSors/EditTaskForm'
 
 const EditWorkOrderTaskPage = ({ query }) => {
   return (
     <>
       <Meta title={`Update Work Order ${query.id}`} />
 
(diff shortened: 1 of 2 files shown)


src/pages/work-orders/[id]/tasks/[taskId]/edit.js
```diff
@@ -0,0 +1,20 @@
...
+import Meta from '../../../../../components/Meta'
+import { getQueryProps } from '../../../../../utils/helpers/serverSideProps'
+import UpdateTaskView from '../../../../../components/WorkOrder/TasksAndSors/UpdateTaskView'
+
+const EditWorkOrderTaskPage = ({ query }) => {
+  return (
+    <>
+      <Meta title={`Update Work Order ${query.id}`} />
+
+      <UpdateTaskView workOrderReference={query.id} taskId={query.taskId} />
```
Comment (siapankina) [inline src/pages/work-orders/[id]/tasks/[taskId]/edit.js:11]: I was thinking about component name 
Maybe we should call it UpdateTaskForm (since it is a form) 
I also have NewTaskView...but have my doubts since it is not just view, it is a form
what do you think ? 
We can rename it later

siapankina (approved):

src/pages/work-orders/[id]/tasks/[taskId]/edit.js
Comment (NeilvB) [reply src/pages/work-orders/[id]/tasks/[taskId]/edit.js]: Good idea. Renamed here https://github.com/LBHackney-IT/repairs-hub-frontend/pull/403/commits/dee5b430d0ef35498181ae8801bb27c4efec92f8

PR #422, opened 2021-11-01
Show operatives on work orders on date that appointment has started

EGiataganas:
### Description of change

Show operatives on work orders on date that appointment has started, before we were debating on the current dateTime, where now only on the date

### DIFF
--- a/src/models/workOrder.test.js
+++ b/src/models/workOrder.test.js
@@ -66,14 +66,14 @@ describe('WorkOrder', () => {
     })
   })
 
-  describe('appointmentStartTimePassed()', () => {
+  describe('appointmentISODatePassed()', () => {
     it('returns false when there is no appointment', () => {
       const workOrder = new WorkOrder()
 
-      expect(workOrder.appointmentStartTimePassed()).toBe(false)
+      expect(workOrder.appointmentISODatePassed()).toBe(false)
     })
 
-    it('returns true when the current time is past the appointment start time', () => {
+    it('returns true when the current date is the same as the appointment start date with past time', () => {
       const now = new Date()
 
       const date = format(now, 'yyyy-MM-dd')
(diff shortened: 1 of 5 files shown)


src/models/workOrder.test.js
```diff
@@ -83,10 +83,10 @@ describe('WorkOrder', () => {
 
       MockDate.set(addMinutes(now, 1))
 
-      expect(workOrder.appointmentStartTimePassed()).toBe(true)
+      expect(workOrder.appointmentISODatePassed()).toBe(true)
     })
 
-    it('returns false when the current time is before the appointment start time', () => {
+    it('returns true when the current time is before the appointment start time', () => {
```
pixeltrix, line 89: Should we have a test for it being false on a before date?

src/models/workOrder.test.js
```diff
@@ -96,7 +96,20 @@ describe('WorkOrder', () => {
...
+      const now = new Date()
+
+      const date = format(now, 'yyyy-MM-dd')
+      const start = format(now, 'HH:mm')
+
+      const workOrder = new WorkOrder({ appointment: { date, start } })
+
+      MockDate.set(subDays(now, 1))
+
+      expect(workOrder.appointmentISODatePassed()).toBe(false)
```
siapankina, line 112: this is duplication of the test above 
could you delete it please ?

src/models/workOrder.test.js
```diff
@@ -96,7 +96,20 @@ describe('WorkOrder', () => {
 
       MockDate.set(subMinutes(now, 1))
 
-      expect(workOrder.appointmentStartTimePassed()).toBe(false)
+      expect(workOrder.appointmentISODatePassed()).toBe(true)
+    })
+
+    it('returns true when the current time is before the appointment start time', () => {
```
siapankina, line 102: since we do, `subDays`
maybe it should be " it returns true when the current day is before the app day"?

pixeltrix, line 102: Should this be `returns false when the current date is before the appointment date` ?

pixeltrix (approved):

PR #430, opened 2021-11-03
Add isRaisable param to sorCodes request

siapankina:
### Description of change
- Added "isRaisable" to API request for SOR codes 
- Updated test for API request (added check is isRaised is true/false)
- Updated integration test for create and update 
- Added new file with "codesWithIsRaisableTrue" as fixture for create and deleted some codes from there 
- For update the fixture is "codes" when all codes can be found

### Further work
- maybe in "cypress/integration/work-order/create/create.spec.js" add test that will check the length of sorCodes that we receive from back end? ("scheduleOfRates/codesWithIsRaisableTrue" length is 5 and "scheduleOfRates/codes" length is 10)
-check why in 'work-order/update.spec.js"  "as('sorCodesRequest')" is not working  (line 80) it fails on lines: 105, 236, 341, 570, 729

### DIFF
--- a/src/utils/frontEndApiClient/scheduleOfRates/codes.test.js
+++ b/src/utils/frontEndApiClient/scheduleOfRates/codes.test.js
@@ -22,6 +22,30 @@ describe('getSorCodes`', () => {
         contractorReference: 'H01',
         propertyReference: '00001234',
         tradeCode: 'PL',
+        isRaisable: false,
+      },
+    })
+  })
+
+  it('calls the Next JS API with isRaisable: true params', async () => {
+    const responseData = { data: 'test' }
+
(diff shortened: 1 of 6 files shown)


src/utils/frontEndApiClient/scheduleOfRates/codes.test.js
```diff
@@ -22,6 +22,30 @@ describe('getSorCodes`', () => {
         contractorReference: 'H01',
         propertyReference: '00001234',
         tradeCode: 'PL',
+        isRaisable: false,
```
EGiataganas, line 25: This can be deleted as the `isRaisable` is an optional param and we will only provide it when is true

EGiataganas:
Closing this in favour #431

PR #441, opened 2021-11-13
Do not create empty variation reasons

siapankina:
https://www.pivotaltracker.com/story/show/180264573


src/utils/hact/workOrderStatusUpdate/updateWorkOrder.js
```diff
@@ -27,9 +27,10 @@ export const buildWorkOrderUpdate = (
     relatedWorkOrderReference: {
       id: reference,
     },
-    ...(variationReason && {
-      comments: `Variation reason: ${variationReason}`,
-    }),
+    ...(variationReason &&
+      variationReason.length != 0 && {
```
EGiataganas, line 31: This can also be just `variationReason.length`, you don't need the `!= 0`

NeilvB (approved):

PR #445, opened 2021-11-17
Add show more/less functionality

siapankina:
https://www.pivotaltracker.com/story/show/180282632


src/components/Layout/Table/index.js
```diff
@@ -43,17 +44,38 @@ export const TH = (props) => (
...
+        props.padding ? `govuk-!-padding-${props.padding}` : null,
+        props.type ? `govuk-table__cell--${props.type}` : null,
+        props.border ? `border-${props.border}` : null,
+        props.className
+      )}
+    >
+      {props.children}
+    </td>
+  )
+}
```
NeilvB, line 81: Now that we have our `<TruncateText >` component we should be able to make the parent responsible for passing that into the `<TD>`, which would leave our original `<TD>` untouched, remove the duplicate code here, and make the parent code easier to read. This should work because the `<TD>` just renders whatever children are passed into it using [`props.children`](https://reactjs.org/docs/composition-vs-inheritance.html#containment).

https://github.com/LBHackney-IT/repairs-hub-frontend/blob/a803af6d0cba0474ea526e88b59d7fd4906bc20b/src/components/WorkOrders/WorkOrderRow.js#L35

should be able to be changed to

```
<TD className="description table-font-size"><TruncateText text={description} numberOfLines="5" /></TD>
```
And then we should be able to revert all changes to `<TD>` in this file.

NeilvB (changes requested):
Just one more step to take with the new component.

src/components/Layout/TruncateText/TruncateText.js
```diff
@@ -0,0 +1,75 @@
...
+}
+
+TruncateText.propTypes = {
+  text: PropTypes.string.isRequired,
+  numberOfLines: PropTypes.string.isRequired,
+  pTagClassName: PropTypes.string,
+  linkClassName: PropTypes.string,
+}
+
+export default TruncateText
```
EGiataganas, line 75: It would be great if this is renamed to `src/components/Layout/TruncateText/index.js`, so later when you want to include it you can do this
```
import TruncateText from '../Layout/TruncateText'
```

src/styles/components/operative/work-order.scss
```diff
@@ -29,13 +29,19 @@
...
   word-break: var(--word-break, 'none');
   overflow: hidden;
 }
+.truncate-line-3 {
+  -webkit-line-clamp: var(--line-clamp, 3);
+}
+
+.truncate-line-5 {
+  -webkit-line-clamp: var(--line-clamp, 5);
+}
```
EGiataganas, line 44: How about
```
.truncate-line {
  &-3 {
     -webkit-line-clamp: var(--line-clamp, 3);
   }

   &-5 {
     -webkit-line-clamp: var(--line-clamp, 5);
   }
 }
 ```

NeilvB: Mm, do you think this version is more readable? I prefer the original. The ampersand trick is neat but here I don't think it adds readability to these tiny utility classes. The number is a bit lost among the other symbols.

siapankina: yes, I agree with Neil 
I had it before with ampersand (not with numbers though), and it seems like having a utility class is more readable

NeilvB (approved):

PR #450, opened 2021-11-23
Fix truncation

NeilvB:


siapankina (approved):
I forgot we had description there too ;)

PR #471, opened 2021-12-10
180394236 select overtime

siapankina:
what is done :

- Checkbox to select overtime the work order page (operative)
- Overtime can be from 4pm-7:59am Mon- Thurs or Fri 4pm- Mon 7:59am (show the checkbox between these times to the operative)
- Bank holidays are also overtime (show the checkbox to the operative then as well)
- When job is closed with Overtime, text will appear: " Work order overtime"

ticket: 
https://www.pivotaltracker.com/story/show/180394236

### DIFF
--- a/src/components/Operatives/OperativeWorkOrder.js
+++ b/src/components/Operatives/OperativeWorkOrder.js
@@ -21,6 +21,7 @@ import OperativeList from './OperativeList'
 import { CLOSED_STATUS_DESCRIPTIONS_FOR_OPERATIVES } from '@/utils/statusCodes'
 import AppointmentHeader from '../WorkOrder/AppointmentHeader'
 import BackButton from '../Layout/BackButton'
+import { isCurrentTimeOperativeOvertime } from '../../utils/helpers/completionDateTimes'
 
 const OperativeWorkOrder = ({
   workOrderReference,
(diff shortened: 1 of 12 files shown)


src/components/Operatives/OperativeWorkOrder.js
```diff
@@ -110,6 +116,19 @@ const OperativeWorkOrder = ({
         readOnly={readOnly}
       />
 
+      {process.env.NEXT_PUBLIC_CAN_CHOOSE_OVERTIME === 'true' &&
+        isCurrentTimeOperativeOvertime() &&
+        !readOnly && (
```
NeilvB, line 121: Good catch

src/styles/components/checkbox.scss
```diff
@@ -0,0 +1,3 @@
+.overtime-checkbox {
+  display: flex !important;
```
NeilvB, line 2: We already have [a class which adds `display: flex`](https://github.com/LBHackney-IT/repairs-hub-frontend/blob/33d0517d26232c8d7332877094d52154e74d96eb/src/styles/layout.scss#L1). Can we use that instead of making a new class and CSS rule or is the `!important` definitely needed?

src/utils/hact/workOrderStatusUpdate/updateWorkOrder.js
```diff
@@ -2,7 +2,8 @@ export const buildWorkOrderUpdate = (
   latestTasks,
   addedTasks,
   reference,
-  variationReason
+  variationReason,
+  isOvertimeChecked
```
NeilvB, line 6: Please can we rename to `isOvertime`. Most of our code dealing with overtime logic shouldn't know about how the user made the selection - if the input element is one day changed to a radio button or select, then this description will be confusing unless we remember to look here and update it. Best to just stick to the core concept in our naming except for the specific component which builds the input.

NeilvB (changes requested):
Requesting we double check if the CSS class is necessary and a small rename.

src/components/Operatives/OperativeWorkOrder.js
```diff
@@ -17,6 +21,7 @@ import OperativeList from './OperativeList'
 import { CLOSED_STATUS_DESCRIPTIONS_FOR_OPERATIVES } from '@/utils/statusCodes'
 import AppointmentHeader from '../WorkOrder/AppointmentHeader'
 import BackButton from '../Layout/BackButton'
+import { isCurrentTimeOperativeOvertime } from '../../utils/helpers/completionDateTimes'
```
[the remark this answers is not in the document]
siapankina: Thank you Neil for pointing out, will have a look

NeilvB (approved):

PR #475, opened 2021-12-14
Update jobStatusUpdate req logic for isOvertime

siapankina:


src/utils/hact/workOrderStatusUpdate/updateWorkOrder.js
```diff
@@ -3,7 +3,8 @@ export const buildWorkOrderUpdate = (
   addedTasks,
   reference,
   variationReason,
-  isOvertime
+  isOvertime,
+  isOvertimeUpdate
```
NeilvB, line 7: Do we need this additional flag? We seem to always set it to `true` when `isOvertime` is supplied. Therefore can't we remove it and just check for the presence of `isOvertime` when building the JSON as we had before?

src/utils/hact/workOrderStatusUpdate/updateWorkOrder.js
```diff
@@ -32,7 +33,7 @@ export const buildWorkOrderUpdate = (
       variationReason.length && {
         comments: variationReason,
       }),
-    ...(isOvertime && { isOvertime: isOvertime }),
+    ...(isOvertimeUpdate && { isOvertime: isOvertime }),
```
NeilvB, line 36: I know it was added in a previous PR but we should add a unit test for this new logic please.

NeilvB (changes requested):

NeilvB (approved):

PR #493, opened 2021-12-21
Components renaming

siapankina:
https://www.pivotaltracker.com/story/show/180705537


NeilvB (approved):
It wasn't in the ticket but please could we do the same treatment on `OperativeWorkOrderView` - I just spotted that one.

Please also move all of these renamed components under `src/components/WorkOrder` or `src/components/WorkOrders` depending on whether they deal with single or multiple WOs.

These can all be in a separate PR.

PR #501, opened 2021-12-22
Update readme with info about environment variables

NeilvB:
https://github.com/LBHackney-IT/repairs-hub-frontend/tree/readme-env-variables#readme


siapankina (approved):
👍

PR #504, opened 2022-01-05
Set overtime when completing

NeilvB:
### Description of change

Avoid overloading the meaning of a "variation request" when updating overtime status on an order. This is resulting in incorrect order statuses.

Instead send this flag to the `workOrderComplete` endpoint.

### Story Link

https://www.pivotaltracker.com/n/projects/2500129/stories/180797615


siapankina (approved):
Noice 😎

src/utils/hact/jobStatusUpdate/variation.js
```diff
@@ -1,10 +1,8 @@
-export const buildWorkOrderUpdate = (
+export const buildVariationFormData = (
```
siapankina, line 1: 👍

PR #505, opened 2022-01-05
180705052 UI improvements

siapankina:
story: https://www.pivotaltracker.com/story/show/180705052

Screenshot 2022-01-05 at 18 43 03
Screenshot 2022-01-05 at 18 43 21
Screenshot 2022-01-05 at 18 43 33

### DIFF
--- a/src/components/Operatives/MobileWorkingWorkOrderDetails.js
+++ b/src/components/Operatives/MobileWorkingWorkOrderDetails.js
@@ -34,18 +34,18 @@ const MobileWorkingWorkOrderDetails = ({
         </h1>
         <div className="priority-text govuk-!-margin-top-0">
           {workOrder.isHigherPriority() ? (
-            <div className="text-dark-red lbh-heading-h3">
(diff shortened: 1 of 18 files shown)


src/components/Operatives/MobileWorkingWorkOrderDetails.js
```diff
@@ -62,82 +55,83 @@ const MobileWorkingWorkOrderDetails = ({
...
+                    {cautionaryAlertsType}
+                  </strong>
+                </div>
+              </GridColumn>
+            </GridRow>
+          )}
 
-        {cautionaryAlertsType && (
-          <GridRow>
+          <GridRow className="align-grid-row">
```
NeilvB, line 91: Best to replace this custom class with just a good old `govuk-!-margin-top-0`. It's normally best to use any utility classes when we want single declarations like this and they are available.

NeilvB (approved):
Looks good - just the class recommendation which can be in a follow-up PR - up to you.

PR #513, opened 2022-01-11
Search and paginate properties via new endpoint

NeilvB:
### Description of change
A [new property search endpoint](https://github.com/LBHackney-IT/repairs-api-dotnet/pull/543) has been added to the service API.

This PR supports that with the old endpoint behind a deprecated flag.

* The conditional logic between endpoints is achieved in an API page route
* Both endpoint responses are wrapped in an object to give our components the same interface
* A Pagination component has been added

### Story Link

https://www.pivotaltracker.com/n/projects/2500129

### UI

image


siapankina (approved):
😎

PR #526, opened 2022-01-17
Add Planned to priority list

siapankina:


src/components/Property/RaiseWorkOrder/SelectPriority.js
```diff
@@ -0,0 +1,63 @@
...
+        register={register({
+          required: 'Please select a priority',
+          validate: (value) =>
+            priorities.map((p) => p.priorityCode).includes(parseInt(value)) ||
+            'Priority is not valid',
+        })}
+        error={errors && errors.priorityCode}
+        widthClass="govuk-!-width-full"
+      />
+      {(drsScheduled || priorityCode == '5') && (
```
siapankina, line 42: Added extra layer in case when trade that user selects already has `priorityCode 5 ==> PLANNED`

src/components/Property/RaiseWorkOrder/SelectPriority.js
```diff
@@ -0,0 +1,63 @@
...
+  register,
+  errors,
+  priorityCode,
+  priorityCodesWithoutDrs,
+}) => {
+  const [drsScheduled, setDrsScheduled] = useState(false)
+
+  const onSelect = (e) => {
+    setDrsScheduled(priorityCodesWithoutDrs.includes(e.target.value))
+    onPrioritySelect(e)
```
NeilvB, line 18: Suggest we change this

```
const selectedCode = parseInt(e.target.value)
setDrsScheduled(priorityCodesWithoutDrs.includes(selectedCode))
onPrioritySelect(selectedCode)
```

That last call would work if the parent just passes `setPriorityCode` straight down as the `onPrioritySelect` prop. Then we should be able to delete `onPrioritySelect` in the parent.

src/components/Property/RaiseWorkOrder/SelectPriority.test.js
```diff
@@ -0,0 +1,75 @@
...
+      <SelectPriority
+        priorities={props.priorities}
+        onPrioritySelect={props.onPrioritySelect}
+        priorityCode="5"
+        priorityCodesWithoutDrs={props.priorityCodesWithoutDrs}
+        register={props.register}
+      />
+    )
+    expect(asFragment()).toMatchSnapshot()
+  })
```
NeilvB, line 74: We should have a test which selects an option and expects to see the info box appear. This could be done within jest or in an integration test.

NeilvB (changes requested):
I've added a few notes but I think it's not too onerous. We'll need to stick with priority code as an integer wherever possible, and add a test. We can also pair up on this.

src/components/Property/RaiseWorkOrder/SelectPriority.js
NeilvB: My earlier code example in Slack while exploring this didn't include integer parsing so that's probably the cause here.

src/utils/helpers/priorities.js
```diff
@@ -2,7 +2,16 @@ export const IMMEDIATE_PRIORITY_CODE = 1
...
 export const URGENT_PRIORITY_CODE = 3
 export const NORMAL_PRIORITY_CODE = 4
+export const PLANNED_PRIORITY_CODE = 9
+export const PRIORITY_CODES_WITHOUT_DRS = [PLANNED_PRIORITY_CODE]
 
-export const LOW_PRIORITY_CODES = [URGENT_PRIORITY_CODE, NORMAL_PRIORITY_CODE]
+export const LOW_PRIORITY_CODES = [
+  URGENT_PRIORITY_CODE,
+  NORMAL_PRIORITY_CODE,
+  PLANNED_PRIORITY_CODE,
```
NeilvB, line 11: Once https://github.com/LBHackney-IT/repairs-hub-frontend/pull/530 is merged, we should be able to remove this. I want us to avoid having to maintain this list every time a priority is added to the backend, and I think that PR should help us - we just use the inverse logic to detect a low priority order.

src/utils/helpers/priorities.js
```diff
@@ -2,11 +2,18 @@ export const IMMEDIATE_PRIORITY_CODE = 1
...
 
 export const HIGH_PRIORITY_CODES = [
   IMMEDIATE_PRIORITY_CODE,
   EMERGENCY_PRIORITY_CODE,
 ]
+export const LOW_PRIORITY_CODES = [
+  URGENT_PRIORITY_CODE,
+  NORMAL_PRIORITY_CODE,
+  PLANNED_PRIORITY_CODE,
+]
```
NeilvB, line 16: Please delete this - you weren't to know but since this was implemented in this branch, develop has been updated so that this const is not needed anywhere any more. (Basically, we list only the high priority codes now and assume everything else is low priority, which means new codes added to the backend shouldn't need frontend updates 🤞 .)

NeilvB (approved):

PR #527, opened 2022-01-17
Hide TMO if tmoName: "London Borough of Hackney"

siapankina:
ticket: https://www.pivotaltracker.com/story/show/180660062

### DIFF
--- a/cypress/integration/property/show.spec.js
+++ b/cypress/integration/property/show.spec.js
@@ -382,6 +382,35 @@ describe('Show property', () => {
     })
   })
 
+  context(
+    'Does not display TMO name if TMO is London Borough of Hackney',
+    () => {
+      beforeEach(() => {
+        // Stub request with property response
+        cy.fixture('properties/propertyWithTmo.json')
+          .then((property) => {
+            property.tmoName = 'London Borough of Hackney'
+            cy.intercept(
+              { method: 'GET', path: '/api/properties/00012345' },
+              { body: property }
+            )
+          })
+          .as('property')
+
+        cy.visit('/properties/00012345')
+        cy.wait('@property')
+      })
+
+      it('shows property address in the heading', () => {
+        cy.get('.lbh-heading-h1').contains('12 Test Street')
+      })
+
+      it('does NOT show TMO details', () => {
+        cy.contains('London Borough of Hackney').should('not.exist')
+      })
+    }
+  )
+
   describe('Out of hours link', () => {
     beforeEach(() => {
       cy.intercept(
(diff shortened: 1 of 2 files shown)


src/components/Property/TenureDetails.js
```diff
@@ -20,7 +20,11 @@ const TenureDetails = ({
       <ul className="lbh-list hackney-property-alerts">
         <Tenure tenure={tenure} canRaiseRepair={canRaiseRepair} />
         <Alerts alerts={locationAlerts} alertType="Address" />
-        <TenureDetail text="TMO" detail={tmoName} />
+        {tmoName == 'London Borough of Hackney' ? (
+          ''
+        ) : (
+          <TenureDetail text="TMO" detail={tmoName} />
+        )}
```
NeilvB, line 27: Please could we extract the string into a const like `TMO_HACKNEY_DEFAULT` and put it at the top of the component with a comment like 'Properties with TMO names set to this value aren't actually TMOs'. This context should reduce confusion for a future visitor.

It's also probably best for readability to shorten the logic by using an `&&` instead of a ternary because then we don't need to return the empty string when we want nothing to render.

Pseudocode

```
{tmoName !== TMO_HACKNEY_DEFAULT && ... }
```

cypress/integration/property/show.spec.js
```diff
@@ -382,6 +382,35 @@ describe('Show property', () => {
...
+          })
+          .as('property')
+
+        cy.visit('/properties/00012345')
+        cy.wait('@property')
+      })
+
+      it('shows property address in the heading', () => {
+        cy.get('.lbh-heading-h1').contains('12 Test Street')
+      })
```
NeilvB, line 406: I don't think we need this test as we've already got it covered on https://github.com/LBHackney-IT/repairs-hub-frontend/pull/527/commits/74bcb156e1ea97b33dcc30d683ecd00e48e599aa#diff-5635bf6f4285741ca6764002d84a6e3bb14623239043e1be10e10a1a70e78e48R373 and we don't expect that to change based on the setup for this test.

NeilvB (changes requested):
Nice job with including the test. I think we have a responsibility to add some more context for this though - please see comment - should be a quick change.

src/components/Property/TenureDetails.js
siapankina: haha, I was actually thinking about using ternary but then decided to go with the classic one :D 
will update everything now ;)

NeilvB (approved):
Nice work

PR #529, opened 2022-01-17
Apply colours to status tags

siapankina:
ticket: https://www.pivotaltracker.com/story/show/180915752

Screenshot 2022-01-17 at 17 24 06

### DIFF
--- a/cypress/integration/property/show.spec.js
+++ b/cypress/integration/property/show.spec.js
@@ -183,7 +183,7 @@ describe('Show property', () => {
           cy.contains('1 Feb 2021')
           cy.contains('11:02')
           cy.contains('DOOR ENTRY ENGINEER - DE')
-          cy.contains('In Progress')
+          cy.contains('In progress')
           cy.contains('The latest repair')
         })
         // Check the last row
--- a/cypress/integration/work-order/index/filter.spec.js
+++ b/cypress/integration/work-order/index/filter.spec.js
@@ -251,7 +251,7 @@ describe('Filter work orders', () => {
         cy.contains('Work complete')
       })
       cy.get('[data-ref=10000030]').within(() => {
-        cy.contains('Variation Pending Approval')
+        cy.contains('Variation pending approval')
       })
       cy.get('[data-ref=10000035]').should('not.exist')
       cy.get('[data-ref=10000040]').should('not.exist')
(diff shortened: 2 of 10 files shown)


NeilvB:
I imagine that around the time when the tags were first added, the dev thought that other tags might be added to the system in other places and they wanted to avoid affecting those. So they only styled the tags that were within a `.work-orders-history-table`.

We've since had confirmation that [the colours should apply to all work order status tags](https://www.pivotaltracker.com/story/show/180915752/comments/229257163) - and I'm pretty confident this is now a stable design decision, especially as it comes from the design system. So I suggest we remove any wrapping div class requirement for styling these so any status tags for work orders can be used in any context in the future and they all get the same styling.

Note that we still have to differentiate work order status tags from task tags (which are used in the approval journey - see `VariationAuthorisationSummary`). Both use `<Status />` so it's a bit fiddly - arguably that should be called `<Tag />`. We could go down the road of adding a `<WorkOrderStatusTag />` and a `<TaskStatusTag />`, which each wrap `<Tag />` but that might be overkill..

I think we can take a "middle path" by adding a new class when we use `<Status />` for a work order status tag.

So in other words, in all places we call `<Status />` with work order text, we also pass in a `'work-order-status'` `className`. Then we can just add a new SCSS file which styles these on this class - and no requirement for a wrapping div.

It would look a bit like this:

```
// applies to tags with both .work-order-status and .lbh-tag but no wrapping class requirement

.work-order-status {
  &.lbh-tag {
    &--status-in-progress {
      background-color: repairs-hub-colour('blue');
      color: repairs-hub-colour('white');
    }
    /* ... all the other status colour styles and the below from mobile working ... */
    &--status-closed,
    &--status-completed {
      background-color: repairs-hub-colour('dark-grey');
    }
  }
}
```

NeilvB:
I think we need to apply the same principle to when we use `<Status />`  in `MobileWorkingWorkOrderListItem`. That will require us to move the colour styles from `/src/styles/components/operative/operative-work-order-list-item.scss` into the new rule you've made. I think the benefit of having all of these in one place outweighs the risk of someone using one of these mobile working statuses in another context and getting an unexpected colour. I think the design should be moving towards consistency, but we can revisit if this becomes a problem.

siapankina:
@NeilvB  thank you for detailed comment, it was much easier for me to follow the logic and make updates!

NeilvB (approved):

PR #530, opened 2022-01-17
Add per-environment dotenv files (part of Sentry addition)

NeilvB:
### Description of change

Allow specification of per-Node-environment vars in separate dotenv files, using [dotenv-flow](https://github.com/kerimdzhanov/dotenv-flow).

We want to differentiate between Node environments so we can activate Sentry for production Node only - this supports that effort.

Does NOT use .env as a fallback committed to git as we already have several team members with secrets in this file. Instead use `.env.defaults` as outlined in the library [docs](https://github.com/kerimdzhanov/dotenv-flow#alternative-defaults-envdefaults).

### Story Link

Part of https://www.pivotaltracker.com/n/projects/2500129/stories/180584037


.env.defaults
```diff
@@ -0,0 +1,50 @@
...
+GSSO_URL=
+HACKNEY_JWT_SECRET=
+GSSO_TOKEN_NAME=token
+
+### Application environment
+NEXT_PUBLIC_ENV_NAME=localdev
+
+### Node environment
+NEXT_PUBLIC_NODE_ENV=development
+NODE_ENV=development
```
pixeltrix, line 17: I had to remove the `NODE_ENV` line in mine because it was affecting the unit tests: https://github.com/LBHackney-IT/bonuscalc-frontend/commit/e15d1664a60c81c6559ecaa191cb363f9f52b482#diff-e9cbb0224c4a3d23a6019ba557e0cd568c1ad5e1582ff1e335fb7d99b7a1055d

pixeltrix (approved):
I had to remove this line otherwise it would set the NODE_ENV to

NeilvB:
This ended up being included with the larger PR to add Sentry - closing

PR #531, opened 2022-01-18
Fix overtime query param detection

NeilvB:
This was causing all mobile orders to be submitted as overtime because
by examining the truthiness of isOvertime query param strings the
flag was always being set to true.

### DIFF
--- a/cypress/integration/work-order/attend/close.spec.js
+++ b/cypress/integration/work-order/attend/close.spec.js
@@ -29,6 +29,7 @@ describe('Closing my own work order', () => {
 
       cy.loginWithOperativeRole()
     })
+
     it('the operative cannot set the work as overtime', () => {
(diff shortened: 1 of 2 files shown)


cypress/integration/work-order/attend/close.spec.js
```diff
@@ -149,6 +150,115 @@ describe('Closing my own work order', () => {
...
+              description: '',
+              allocatedBy: '',
+            },
+            jobStatusUpdates: [
+              {
+                typeCode: '0',
+                otherType: 'complete',
+                comments: 'Work order closed - I attended',
+                eventTime: new Date(now.setHours(16, 0, 1)).toISOString(),
+                isOvertime: false,
```
NeilvB, line 244: This is the important part of this test.

pixeltrix (approved):

PR #536, opened 2022-01-19
Update repsonse from api/properties/propertyRef

siapankina:
ticket: https://www.pivotaltracker.com/story/show/180886400

### DIFF
--- a/src/pages/api/properties/[id].test.js
+++ b/src/pages/api/properties/[id].test.js
@@ -28,20 +28,16 @@ describe('GET /api/properties/[id] contact information redaction', () => {
           property: {},
           alerts: {},
           tenure: {},
-          contacts: {
-            contacts: [
-              {
-                firstName: 'FirstName',
-                lastName: 'LastName',
-                phoneNumbers: ['0123456789', '0123456789'],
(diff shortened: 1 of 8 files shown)


NeilvB (changes requested):
What's here looks great but there's something missing which is a real hidden "gotcha" that would have been very difficult to detect if you didn't know it was there.

We need to update a custom API endpoint on the Node server which [anonymises contact details](https://github.com/LBHackney-IT/repairs-hub-frontend/blob/develop/src/pages/api/properties/%5Bid%5D.js#L20) for a particular set of users. This should just be a case of updating the field and some updates to the setup and expectations of its test.

src/pages/api/properties/[id].test.js
```diff
@@ -28,16 +28,14 @@ describe('GET /api/properties/[id] contact information redaction', () => {
           property: {},
           alerts: {},
           tenure: {},
-          contacts: {
-            contacts: [
+          contactDetails: {
```
Comment (NeilvB) [inline src/pages/api/properties/[id].test.js:31]: Shouldn't this be an array?

NeilvB (approved):

PR #537, opened 2022-01-25
Avoid managing two priority form fields

NeilvB:
### Description of change

Previously the priority select part of the form had both its value and text value set to the priority description we get from the API for each priority.

This required some logic and a hidden field to track both the selected description and its code. By levereging the ability of the select to hold a text value different from its internal value (description and code respectively), we can simplify some of this form and its state.

### Story Link

This is groundwork which should hopefully make https://www.pivotaltracker.com/n/projects/2500129/stories/180645774 more straightforward.

### DIFF
--- a/cypress/integration/work-order/create/create.spec.js
+++ b/cypress/integration/work-order/create/create.spec.js
@@ -89,11 +89,9 @@ describe('Raise repair form', () => {
       cy.contains('Please enter a quantity')
     })
 
-    cy.get('#priorityDescription-form-group .govuk-error-message').within(
-      () => {
-        cy.contains('Please select a priority')
-      }
-    )
+    cy.get('#priorityCode-form-group .govuk-error-message').within(() => {
+      cy.contains('Please select a priority')
+    })
 
     cy.get('#descriptionOfWork-form-group .govuk-error-message').within(() => {
       cy.contains('Please enter a repair description')
@@ -170,11 +168,11 @@ describe('Raise repair form', () => {
         'not.be.disabled'
       )
       // Priority disabled until SOR is selected
-      cy.get('#priorityDescription').should('be.disabled')
+      cy.get('#priorityCode').should('be.disabled')
       cy.get('input[id="rateScheduleItems[0][code]"]').type(
         'INP5R001 - Pre insp of wrks by Constructr'
       )
-      cy.get('#priorityDescription').should('not.be.disabled')
+      cy.get('#priorityCode').should('not.be.disabled')
 
       // Selecting no trade clears contractor and SOR code select options
       cy.get('#trade').clear()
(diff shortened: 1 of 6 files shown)


src/components/Property/RaiseWorkOrder/RaiseWorkOrderForm.js
```diff
@@ -28,32 +28,32 @@ const RaiseWorkOrderForm = ({
...
 
   const [totalCost, setTotalCost] = useState('')
   const overSpendLimit = totalCost > raiseLimit
 
   const onSubmit = async (formData) => {
-    const scheduleWorkOrderFormData = buildScheduleWorkOrderFormData(formData)
+    const scheduleWorkOrderFormData = buildScheduleWorkOrderFormData({
+      ...formData,
+      priorityDescription: getPriorityObjectByCode(formData.priorityCode)
+        .description,
```
NeilvB, line 41: Needed because the form no longer has any field that contains the description - it's derived from the code, now.

src/components/Property/RaiseWorkOrder/RaiseWorkOrderForm.js
```diff
@@ -67,30 +67,31 @@ const RaiseWorkOrderForm = ({
...
+    if (getPriorityObjectByDescription(description)) {
       // Update priority when SOR code has priority attached if:
       // Priority description is blank, or there's only one sor code entry, or
       // when removing an SOR there's an existing entry with higher priority, or
       // the selected priority code is less than existing priority codes
       // (Higher priority as code gets lower)
+
       if (
-        !existingCode ||
+        !priorityCode ||
```
NeilvB, line 78: Updating the select (or doing it behind-the-scenes via SOR code selection) now results in an update to the `priorityCode`, so we no longer need to query the DOM for it.

src/components/Property/RaiseWorkOrder/RaiseWorkOrderForm.js
```diff
@@ -67,30 +67,31 @@ const RaiseWorkOrderForm = ({
...
-        document.getElementById('priorityDescription').value = description
-
         setPriorityCode(
           getPriorityObjectByDescription(description)?.priorityCode
         )
+
+        setValue(
+          'priorityCode',
+          getPriorityObjectByDescription(description)?.priorityCode
+        )
```
NeilvB, line 94: Without this, the priority code state update is not reflected in the select. I'm not sure right now if there's a better way - I'm sure there's a technique available in react-hook-form but I haven't found it yet.

src/components/Property/RaiseWorkOrder/RateScheduleItemView.js
```diff
@@ -70,7 +70,7 @@ const RateScheduleItemView = ({
   }
 
   const onRateScheduleItemSelect = (index, event) => {
-    document.getElementById('priorityDescription').disabled = false
+    document.getElementById('priorityCode').disabled = false
```
NeilvB, line 73: This is a kind of global side effect of this component (this will error if reused in another context without a priority code somewhere on the page) but I wasn't sure of a solution at the moment. Ideally the parent would know about the SOR change from its own state and then set a flag on the priority select. Might be one for another time.

cypress/integration/work-order/create/create.spec.js
```diff
@@ -203,16 +201,18 @@ describe('Raise repair form', () => {
...
+      cy.get('#priorityCode').should('have.value', '')
 
       // Select SOR code with priority attached
       cy.get('input[id="rateScheduleItems[0][code]"]')
         .clear()
         .type('DES5R003 - Immediate call outs')
+
       // Autopopulates priority description
-      cy.get('#priorityDescription').should('have.value', '1 [I] IMMEDIATE')
+      cy.get('#priorityCode')
```
siapankina, line 212: shouldn't we check here if `#priorityCode` has value '`1 [I] IMMEDIATE'`
something like : `cy.get('#priorityCode')
        .find('option:selected')
        .should('have.text', '1 [I] IMMEDIATE')`

siapankina (approved):

cypress/integration/work-order/create/create.spec.js
NeilvB: 👍

PR #538, opened 2022-01-26
Add a proper sign out test

NeilvB:
Previously we were manipulating the cookie within the test setup rather than properly testing the application sign out logic.


package.json
```diff
@@ -11,13 +11,13 @@
...
+    "test:unit": "NEXT_PUBLIC_ENV_NAME=test node --expose-gc ./node_modules/.bin/jest --runInBand --logHeapUsage",
     "lint": "eslint . --max-warnings 0 --ext .js,.jsx && echo 'Lint complete.'",
-    "test:e2e": "PORT=5001 start-test 5001 cy:run",
-    "e2e:audit": "PORT=5001 start-test 5001 cy:run:audit",
-    "e2e:interactive": "PORT=5001 start-test 5001 cy:open",
-    "e2e:server": "PORT=5001 start-test 5001",
+    "test:e2e": "NEXT_PUBLIC_ENV_NAME=test PORT=5001 start-test 5001 cy:run",
+    "e2e:audit": "NEXT_PUBLIC_ENV_NAME=test PORT=5001 start-test 5001 cy:run:audit",
+    "e2e:interactive": "NEXT_PUBLIC_ENV_NAME=test PORT=5001 start-test 5001 cy:open",
+    "e2e:server": "NEXT_PUBLIC_ENV_NAME=test PORT=5001 start-test 5001",
```
NeilvB, line 20: These will all be derived from `NODE_ENV` once we switch to dotenv-flow. There is an open PR for that but it will need updating.

Asura5: Do you need && before jest to execute the shell param store or is that not needed?

NeilvB: I don't think it's needed - just checked and code executed from Jest can access this value.

Asura5 (approved):

teodor-hristoforov-hackney (approved):

NeilvB:
Thanks for the help with reviews on this.

PR #542, opened 2022-01-27
Handle new priorities by removing unnecessary code-time mappings

NeilvB:
### Description of change

The frontend has long held its own mappings between priority codes and the expected target number of days and hours associated with that code. It doesn't need to do this.

This is an issue for a few reasons including when more priorities are added to the API DB as they don't work with a matching update to the frontend mapping.

This PR removes the mapping by doing the following:
 - Taking the target days from the priority object returned from the API
 - Deriving the target hours from the above (except for immediates which is hardcoded as 2 hours)
 - Reworking the completion time calculation function to take days and hours parameters instead of a code that is used to look up the associated times in the mapping.

The last point meant our unit tests for this function were restructured with some clearer contexts.

### Story Link

https://www.pivotaltracker.com/n/projects/2500129/stories/180293385


teodor-hristoforov-hackney (approved):

src/components/Property/RaiseWorkOrder/RaiseWorkOrderForm.js
```diff
@@ -13,6 +13,8 @@ import TradeContractorRateScheduleItemView from './TradeContractorRateScheduleIt
 import Contacts from '../Contacts/Contacts'
 import WarningText from '../../Template/WarningText'
 import { buildScheduleWorkOrderFormData } from '@/utils/hact/workOrderSchedule/raiseWorkOrderForm'
+import { IMMEDIATE_PRIORITY_CODE } from '../../../utils/helpers/priorities'
```
siapankina, line 16: just a nitpick , sorry for that 
we can use '@/utils....' like the way you do in `raiseWorkOrderForm.js` to stay consistent

siapankina (approved):

PR #557, opened 2022-02-02
Show support link for operatives

siapankina:
Ticket: https://www.pivotaltracker.com/story/show/180993602
Since we can't use the same logic we did for PhaseBanner (check if user `canAttendOwnWorkOrder` ==> user is null/undefined on login page) I decided to use custom css to hide the text when user is on tablet /screen view

Login page:
Screenshot 2022-02-02 at 16 50 06

Mobile menu:
Screenshot 2022-02-02 at 16 50 35


src/components/UserLogin/index.js
```diff
@@ -30,6 +31,13 @@ const UserLogin = ({ gssoUrl, submitText }) => {
         <p className="lbh-body">
           Please contact your manager if you have issues signing in.
         </p>
+
+        <p className="lbh-body text-on-mobile-view">
```
NeilvB, line 35: I would name this class based on something to do with its content, like `support-text`or `support-link`.  That way, if the design changes to show this on eg tablet as well as mobile, then we don't have to remember to update it to keep the readability intact.

src/utils/headerLinks.js
```diff
@@ -29,13 +29,20 @@ const HEADER_LINKS = [
...
   {
-    href: 'work-orders/cautionary-alerts',
+    href: '/work-orders/cautionary-alerts',
     id: 'cautionary-alerts',
     description: 'Cautionary Alerts',
     permittedRoles: [OPERATIVE_ROLE],
   },
+
+  {
+    href: 'https://sites.google.com/hackney.gov.uk/repairshubsupport',
```
NeilvB, line 39: Ideally this would sit in one of the `NEXT_PUBLIC_` environment variables. I don't think this prevents a merge though.

NeilvB (dismissed):
It's a bit picky of me but I'm requesting we update the classname.

Do you think we'd benefit from adding a check for the new link in [this Cypress test](https://github.com/LBHackney-IT/repairs-hub-frontend/blob/develop/cypress/integration/navigation.spec.js#L23)?

pixeltrix (approved):

PR #558, opened 2022-02-03
Planned priority is now 365 calendar days

siapankina:
Planned priority is now calculated as 365 calendar days 
+ updated warning text as design requested

Screenshot 2022-02-03 at 14 03 34

### DIFF
--- a/src/utils/hact/workOrderSchedule/raiseWorkOrderForm.js
+++ b/src/utils/hact/workOrderSchedule/raiseWorkOrderForm.js
@@ -19,6 +19,7 @@ export const buildScheduleWorkOrderFormData = (workOrderData) => {
         lowPriority: !HIGH_PRIORITY_CODES.includes(
           Number.parseInt(workOrderData.priorityCode)
         ),
+        plannedPriority: Number.parseInt(workOrderData.priorityCode) == 9,
       }),
       numberOfDays: workOrderData.daysToComplete,
     },
--- a/src/utils/hact/workOrderSchedule/raiseWorkOrderForm.js
+++ b/src/utils/hact/workOrderSchedule/raiseWorkOrderForm.js
@@ -1,6 +1,9 @@
 import { v4 as uuidv4 } from 'uuid'
 import { calculateCompletionDateTime } from '../../helpers/completionDateTimes'
-import { HIGH_PRIORITY_CODES } from '@/utils/helpers/priorities'
+import {
+  HIGH_PRIORITY_CODES,
+  PLANNED_PRIORITY_CODE,
+} from '@/utils/helpers/priorities'
 
 export const buildScheduleWorkOrderFormData = (workOrderData) => {
   return {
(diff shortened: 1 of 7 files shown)


src/utils/hact/workOrderSchedule/raiseWorkOrderForm.js
```diff
@@ -19,6 +19,7 @@ export const buildScheduleWorkOrderFormData = (workOrderData) => {
         lowPriority: !HIGH_PRIORITY_CODES.includes(
           Number.parseInt(workOrderData.priorityCode)
         ),
+        plannedPriority: Number.parseInt(workOrderData.priorityCode) == 9,
```
NeilvB, line 22: Use the constant we've defined in `priorities.js` for this, with a `===`

src/utils/helpers/completionDateTimes.js
```diff
@@ -78,12 +78,15 @@ export const calculateCompletionDateTime = ({
   workingDays,
   workingHours = 0,
   lowPriority = false,
+  plannedPriority = false,
```
NeilvB, line 81: Please write a Jest test for this

NeilvB (changes requested):

src/utils/hact/workOrderSchedule/raiseWorkOrderForm.js
```diff
@@ -19,7 +19,7 @@ export const buildScheduleWorkOrderFormData = (workOrderData) => {
         lowPriority: !HIGH_PRIORITY_CODES.includes(
           Number.parseInt(workOrderData.priorityCode)
         ),
-        plannedPriority: Number.parseInt(workOrderData.priorityCode) == 9,
+        plannedPriority: Number.parseInt(workOrderData.priorityCode) === 9,
```
NeilvB, line 22: We need to use `PLANNED_PRIORITY_CODE` here

src/utils/hact/workOrderSchedule/raiseWorkOrderForm.test.js
```diff
@@ -162,4 +188,95 @@ describe('buildScheduleWorkOrderFormData', () => {
     const response = buildScheduleWorkOrderFormData(workOrderData)
     expect(response).toEqual(scheduleWorkOrderFormData)
   })
+
+  it('builds the ScheduleWorkOrder form data to post to the Repairs API with Planned Priority', async () => {
```
NeilvB, line 192: 👍  This is a good test to add, which I didn't think of. However we will still benefit from a test at the unit level in this file `completionDateTimes.test.js`. Sorry - I wasn't clear enough in my earlier comment.

You can just add another `describe` at the top level e.g. `'when the plannedPriority is true'`. I think it's important that these target times are put through their paces so having the additional coverage should be worth it.

NeilvB (changes requested):

src/utils/hact/workOrderSchedule/raiseWorkOrderForm.js
siapankina: ah, sorry for that

pixeltrix (approved):

PR #559, opened 2022-02-03
Design changes follow up pr

siapankina:
Addressed design comments from this ticket --> https://www.pivotaltracker.com/story/show/180915752
What was addressed: 
- `status options on the left hand side to also be "Sentence case" (in filters section)`
- `The "Completed" tag should have a black border and a white background on both RH and mobile`
------------------------------------------------------------------------------------------------------

`Can we remove "Work" from "Work cancelled" and change "Complete" to "Completed" in both the table results and the left hand side option and on mobile.`Will addressed in separate PR after b/e makes changes

------------------------------------------------------------------------------------------------------

Screenshot:

Screenshot 2022-02-04 at 12 37 10

### DIFF
--- a/src/utils/helpers/textConverter.js
+++ b/src/utils/helpers/textConverter.js
@@ -1,0 +1,3 @@
+export const convertToSentenceCases = (text) => {
+  return text.charAt(0) + text.slice(1).toLocaleLowerCase()
+}
--- a/src/utils/helpers/textConverter.test.js
+++ b/src/utils/helpers/textConverter.test.js
@@ -1,0 +1,5 @@
+import { convertToSentenceCases } from './textConverter'
+
+it('Converts text into sentence case', () => {
+  expect(convertToSentenceCases('HELLO WORLD')).toEqual('Hello world')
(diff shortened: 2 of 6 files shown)


src/utils/helpers/textConverter.js
```diff
@@ -0,0 +1,3 @@
+export const convertToSentenceCases = (text) => {
+  return text.charAt(0) + text.slice(1).toLocaleLowerCase()
+}
```
siapankina, line 3: I actually can add test to this function :D 
will add it now

siapankina: ✅

src/utils/helpers/textConverter.test.js
```diff
@@ -0,0 +1,5 @@
+import { convertToSentenceCases } from './textConverter'
+
+it('Converts text into sentece case', () => {
```
pixeltrix, line 3: Typo in the test description

pixeltrix (approved):

PR #564, opened 2022-02-07
Don't show appointment link if appointment not req

siapankina:


src/models/workOrder.js
```diff
@@ -22,8 +23,13 @@ export class WorkOrder {
     )
   }
 
+  isAppointmentRequired = () => {
+    console.log(this.priorityCode)
```
NeilvB, line 27: Looks like some test logging which can be deleted

src/components/WorkOrder/WorkOrderHeader.test.js
```diff
@@ -15,7 +15,7 @@ describe('WorkOrderHeader component', () => {
     description: 'This is an urgent repair description',
     propertyReference: '00014888',
     status: 'Work Complete',
-    priorityCode: 'some code',
+    priorityCode: 3,
```
NeilvB, line 18: We should probably use the `URGENT_PRIORITY_CODE` const here to save a reader looking up what the number means.

NeilvB (approved):

PR #573, opened 2022-02-10
Open support link in new tab

siapankina:
ticket: https://www.pivotaltracker.com/story/show/181204123


NeilvB:
Please add a test in e.g. `navigation.spec.js` for this. Chaining on `.should('have.attr', 'target', '_blank')` should do it.

NeilvB (approved):

PR #578, opened 2022-02-16
Semi automated filters

siapankina:
With this PR use can remove selected filters manually one by one 

ticket: https://www.pivotaltracker.com/story/show/180915323
Screenshot 2022-02-16 at 17 41 46

### DIFF
--- a/cypress/integration/work-order/index/filter.spec.js
+++ b/cypress/integration/work-order/index/filter.spec.js
@@ -464,6 +464,72 @@ describe('Filter work orders', () => {
       cy.get('[data-ref=10000032]')
     })
 
+    it('Clears filters one by one', () => {
+      cy.wait(['@filters', '@workOrders'])
+      //SElect Status filters
+      cy.get('.govuk-checkboxes').find('[name="StatusCode.90"]').check()
+      cy.get('.govuk-checkboxes').find('[name="StatusCode.80"]').check()
+      //Select Priority filters
+      cy.get('.govuk-checkboxes').find('[name="Priorities.2"]').check()
+
+      cy.get('[data-testid=apply-filters]').contains('Apply filters').click()
+
+      cy.wait([
+        '@filters',
+        '@workOrdersInProgressVariationPendingApprovalEmergency',
+      ])
+
+      cy.get('.selected-filters').within(() => {
+        cy.get('#selected-filters-Status').within(() => {
+          cy.contains('Variation Pending Approval')
+          cy.contains('In Progress')
+        })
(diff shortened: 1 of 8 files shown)


src/components/WorkOrders/WorkOrdersView.js
```diff
@@ -38,6 +39,31 @@ const WorkOrdersView = ({ query }) => {
...
+      case 'status':
+        updatedAppliedFilters.StatusCode.splice(indexNumber, 1)
+        break
+      case 'trade':
+        updatedAppliedFilters.TradeCodes.splice(indexNumber, 1)
+        break
+    }
+
+    updateUrlQueryParams(updatedAppliedFilters)
+    delete updatedAppliedFilters['pageNumber']
```
NeilvB, line 63: Why do we need to do this?

cypress/integration/work-order/index/filter.spec.js
```diff
@@ -464,6 +464,88 @@ describe('Filter work orders', () => {
...
+      cy.get('.selected-filters').within(() => {
+        cy.get('#selected-filters-Status').within(() => {
+          cy.contains('li', 'Variation Pending Approval').within(() => {
+            cy.get('button')
+              .should(
+                'have.attr',
+                'aria-label',
+                'Remove Variation Pending Approval filter'
+              )
+              .click()
```
NeilvB, line 504: When we remove filters in this test, do any of them result in real API calls? An easy way to check is to turn off your connection, or to look at your logs - real API requests should show `Starting API Request` or similar. If so we need to create stubbed responses with `wait` calls, in the same way as for other requests in this file.

NeilvB: We shouldn't need to bother with actually returning any work orders for these additional stubs - we're not asserting anything on the data that's returned. We can just base them on this, replacing the path and name as appropriate.

```
    cy.intercept(
      {
        path:
          '/api/workOrders/?PageSize=10&PageNumber=1&StatusCode=30&IncludeHistorical=false',
      },
      { body: [] }
    ).as('workOrdersCancelled')
```

src/components/WorkOrders/WorkOrdersView.js
siapankina: I was trying to stay consistent with the logic of passing the same values (https://github.com/LBHackney-IT/repairs-hub-frontend/blob/a3cf733ca6ebd8cc28bf3c4bd751f310c1997389/src/components/WorkOrders/WorkOrdersView.js#L25) 
to `setAppliedFilters` we always pass object of arrays with filter values 
thats the only reason

PR #602, opened 2022-03-03
Allow updating WOs with multi trades

siapankina:


NeilvB (approved):
Thanks for adding the flag here 👍

PR #634, opened 2022-03-25
Refactor success confirmation pages

siapankina:
ticket: https://www.pivotaltracker.com/story/show/181657598
Refactored all (at least what I found) confirmation pages for : Close, Update, Cancel and Variation actions 
instead of using several components we will have just 1 component with switch case


NeilvB:
I like the overall thinking here, but as discussed in Slack I think we should hold off until we find out if design want to add more consistency to success pages. If they do, this could become quite a bit simpler.

pixeltrix (approved):
I assume that design gave the 👍🏻 in relation to Neil's comment?

NeilvB (approved):

PR #646, opened 2022-04-04
Configure JS excludes for VSCode

NeilvB:
Exclude directories such as builds from JS analysis so that VSCode stops complaining. Should make "Intellisense" e.g. cmd-click function navigation more reliable. See https://code.visualstudio.com/docs/languages/jsconfig#_using-the-exclude-property


siapankina (approved):
😎

PR #655, opened 2022-04-20
 Variation authorisation Purdy

siapankina:
Ticket: https://www.pivotaltracker.com/story/show/181525325

Screenshot 2022-04-20 at 15 01 03

Screenshot 2022-04-20 at 15 01 24

### DIFF
--- a/cypress/integration/work-order/authorisation/variation-authorisation.spec.js
+++ b/cypress/integration/work-order/authorisation/variation-authorisation.spec.js
@@ -361,13 +361,14 @@ describe('Contract manager can authorise variation', () => {
     })
   })
 
-  it('Can reject variation for Purdy contractor wirh unlimited characters for notes and will show summary page before submitting rejection', () => {
+  it('Can reject variation for Purdy contractor with unlimited characters for notes and will show summary page before submitting rejection', () => {
     cy.intercept(
       { method: 'GET', path: '/api/workOrders/10000012' },
       { fixture: 'workOrders/statusVariationPendingApprovalPurdy.json' }
-    )
+    ).as('workOrderPurdy')
 
     cy.visit('/work-orders/10000012/variation-authorisation')
+    cy.wait('@workOrderPurdy')
 
     cy.contains('Authorisation variation request: 10000012')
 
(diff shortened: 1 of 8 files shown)


cypress/integration/work-order/authorisation/variation-authorisation.spec.js
```diff
@@ -360,4 +360,72 @@ describe('Contract manager can authorise variation', () => {
       cy.contains('td', '£10')
     })
   })
+
+  it('Can reject variation for Purdy contractor wirh unlimited characters for notes and will show summary page before submitting rejection', () => {
```
NeilvB: We can rework this a bit to avoid repeating the long string. At the moment it's tricky to see if the string changes or if it's just the same thing over and over.

We can define the var like this with the relevant string length passed into `.repeat()` - also avoids the need for any specific copy and surfaces the length which is the important bit - also means we don't need a comment about the length.

```
  cy.wrap('x'.repeat(251)).as('longString')
```

Then add a callback and place everything that needs to refer to the string within it:

```
    cy.get('@longString').then((longString) => {
      cy.get('#note').type(longString)
      // other commands
      cy.contains(`This work is too complicated: ${longString}`)  // interpolated into another string
      // other commands
    })
```

src/components/WorkOrder/Authorisation/VariationAuthorisationView.js
```diff
@@ -181,34 +202,72 @@ const VariationAuthorisationView = ({ workOrderReference }) => {
...
+                    <>
+                      <h3 className="lbh-heading-h3">
+                        You are rejecting the variation request
+                      </h3>
+                      <div className="lbh-stat">
+                        <span className="lbh-stat__caption">
+                          {rejectionReasonToShow}
+                          <Link href="#">
+                            <a
+                              onClick={changeStatus}
```
NeilvB, line 257: Please can we give this a more specific name like `showEditPage` or similar.

cypress/integration/work-order/authorisation/variation-authorisation.spec.js
```diff
@@ -360,4 +360,60 @@ describe('Contract manager can authorise variation', () => {
...
+    cy.wait('@apiCheck', { requestTimeout: 6000 })
+
+    cy.get('@apiCheck')
+      .its('request.body')
+      .should('deep.equal', {
+        relatedWorkOrderReference: {
+          id: '10000012',
+        },
+        comments:
+          'Variation rejected: This work is too complicated: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx',
```
NeilvB, line 404: If you put this and all above commands inside the `cy.get('@longString')...` callback then we never need to hard-code the full 251 char string.

NeilvB (approved):

PR #656, opened 2022-04-21
Vary Purdy with incremental search

NeilvB:
### Description

Allow Purdy variations to take place across all of Purdy's SORs. Uses the same incremental search approach as Purdy raising.

### Story

https://www.pivotaltracker.com/n/projects/2500129/stories/181508986

### UI

image

### DIFF
--- a/src/components/WorkOrder/Update/index.js
+++ b/src/components/WorkOrder/Update/index.js
@@ -9,6 +9,7 @@ import { isSpendLimitReachedResponse } from '@/utils/helpers/apiResponses'
 import WorkOrderUpdateForm from './Form'
 import WorkOrderUpdateSummary from './Summary'
 import WorkOrderUpdateSuccess from './Success'
+import { PURDY_CONTRACTOR_REFERENCE } from '@/utils/constants'
 
 const WorkOrderUpdateView = ({ reference }) => {
   const [loading, setLoading] = useState(false)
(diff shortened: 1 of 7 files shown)


src/components/WorkOrder/Update/index.js
```diff
@@ -85,35 +130,48 @@ const WorkOrderUpdateView = ({ reference }) => {
...
-          propertyReference: workOrder.propertyReference,
-          contractorReference: workOrder.contractorReference,
-          showAdditionalTrades: true,
-        },
-      })
+      const multiTradeIncrementalSearch = await incrementalSORSearchRequired(
+        workOrder.contractorReference
+      )
 
+      console.log('multiTradeIncrementalSearch', multiTradeIncrementalSearch)
```
siapankina, line 148: probably can get rid of that

src/components/WorkOrder/Update/index.js
```diff
@@ -9,14 +9,15 @@ import { isSpendLimitReachedResponse } from '@/utils/helpers/apiResponses'
...
+import { PURDY_CONTRACTOR_REFERENCE } from '@/utils/constants'
 
 const WorkOrderUpdateView = ({ reference }) => {
   const [loading, setLoading] = useState(false)
   const [error, setError] = useState()
   const [currentUser, setCurrentUser] = useState({})
   const [tasks, setTasks] = useState([])
   const [originalTasks, setOriginalTasks] = useState([])
-  const [sorCodes, setSorCodes] = useState([])
+  const [workOrder, setWorkOrder] = useState([])
```
siapankina, line 20: Is there a particular reason why we need to use `array[]` instead of `object{}`?
Cause from b/e WorkOrder comes as an object
`{reference: 10000954, dateRaised: '2022-04-21T10:54:48.394808Z',.....}`

NeilvB: No :) good spot - updated to use `useState()`

siapankina (approved):

PR #660, opened 2022-04-25
Assigning correct budget code

siapankina:
ticket: https://www.pivotaltracker.com/story/show/181446341

Right now budget code can be assigned to all contractors
- if chosen Purdy then only specific budget codes will be shown 
- if chosen any other contractor all budget codes are shown
- only users that has permission to assign budget codes will be able to assign budget codes


src/components/Property/RaiseWorkOrder/RaiseWorkOrderForm.test.js
```diff
@@ -102,4 +103,31 @@ describe('RaiseWorkOrderForm component', () => {
 
     expect(asFragment()).toMatchSnapshot()
   })
+
+  it('should render without possibility to choose budget code', async () => {
```
NeilvB, line 107: Nitpick but could we add info on what controls this to the description, e.g. `...for a user without the budget code officer permission`.

NeilvB (approved):

PR #665, opened 2022-04-27
Upload multiple sor codes [skeleton]

siapankina:
Skeleton for uploading multiple SORs ticket: https://www.pivotaltracker.com/story/show/181730893

- add link for vary and raise repair form 
- add new component where multiple SORs can be added 
- refactor code (page steps) for vary and raise 
- add integration tests (click on the link , see new page) 
- update snapshots 

Screenshot 2022-04-28 at 17 51 18
Screenshot 2022-04-28 at 17 54 22

### DIFF
--- a/cypress/integration/work-order/create/create-with-adding-multiple-sors.spec.js
+++ b/cypress/integration/work-order/create/create-with-adding-multiple-sors.spec.js
@@ -1,0 +1,92 @@
+const now = new Date('2022-02-11T12:00:00')
+describe('Raise repair form', () => {
+  beforeEach(() => {
+    cy.intercept(
+      { method: 'GET', path: '/api/properties/00012345' },
+      { fixture: 'properties/property.json' }
+    ).as('propertyRequest')
+
+    cy.intercept(
+      {
+        method: 'GET',
+        path: '/api/contractors?propertyReference=00012345&tradeCode=PL',
+      },
+      { fixture: 'contractors/contractors.json' }
+    ).as('contractorsRequest')
+
+    cy.intercept(
+      { method: 'GET', path: '/api/schedule-of-rates/priorities' },
+      { fixture: 'scheduleOfRates/priorities.json' }
+    ).as('sorPrioritiesRequest')
+
+    cy.intercept(
+      { method: 'GET', path: '/api/schedule-of-rates/trades?propRef=00012345' },
+      { fixture: 'scheduleOfRates/trades.json' }
+    ).as('tradesRequest')
+
+    cy.intercept(
+      {
+        method: 'GET',
+        path: '/api/workOrders?*',
+      },
+      { body: [] }
+    ).as('workOrdersRequest')
+
+    cy.intercept(
+      {
+        method: 'GET',
+        path: '/api/workOrders/budget-codes?contractorReference=PCL',
+      },
+      { fixture: 'scheduleOfRates/budgetCodes.json' }
+    ).as('budgetCodesRequest')
+
+    cy.intercept(
+      {
+        method: 'GET',
+        path:
+          '/api/schedule-of-rates/codes?tradeCode=PL&propertyReference=00012345&contractorReference=PCL&isRaisable=true',
+      },
+      { fixture: 'scheduleOfRates/codesWithIsRaisableTrue.json' }
+    ).as('sorCodesRequest')
+
+    cy.intercept(
+      { method: 'GET', path: '/api/toggles' },
+      {
+        body: [
+          {
+            featureToggles: { MultiTradeSORIncrementalSearch: true },
+          },
+        ],
+      }
+    ).as('featureToggle')
+
+    cy.clock(now, ['Date'])
+  })
+  context('When choosing to add multiple SOR codes option', () => {
+    beforeEach(() => {
+      cy.loginWithAgentRole()
+    })
+
+    it('allows to add multiple SORs', () => {
+      cy.visit('/properties/00012345/raise-repair/new')
+
+      cy.wait(['@propertyRequest', '@sorPrioritiesRequest', '@tradesRequest'])
+
+      cy.get('#repair-request-form').within(() => {
+        cy.get('#trade').type('Plumbing - PL')
+
+        cy.wait('@contractorsRequest')
+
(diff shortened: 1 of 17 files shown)


cypress/integration/work-order/update-with-adding-multiple-sors.spec.js
```diff
@@ -0,0 +1,54 @@
+describe('Updating a work order', () => {
```
NeilvB, line 1: We've got two update specs now so I'd say it's time to make a directory like we have for `create` for this and the existing update spec, under `cypress/integration/work-order/update/`

cypress/integration/work-order/update.spec.js
```diff
@@ -326,7 +326,7 @@ describe('Updating a work order', () => {
       })
 
       cy.get('#repair-request-form').within(() => {
-        cy.get('.lbh-link').click()
+        cy.contains('+ Add another SOR').click()
```
NeilvB, line 329: Good test improvement

src/components/Property/RaiseWorkOrder/AddMultipleSORs.js
```diff
@@ -0,0 +1,39 @@
...
+      <form
+        role="form"
+        id="adding-multiple-sors-form"
+        onSubmit={handleSubmit(onSubmit)}
+      >
+        <TextArea
+          name="note"
+          label="Enter SOR codes as a list:"
+          register={register({
+            required: 'Please add notes',
```
NeilvB, line 29: This message should reference SOR codes. There's no specified message in the ticket but I suggest we use 'Please enter SOR codes'. I didn't mention it earlier but please could we also include a check in the Cypress tests for the presence validation message after submitting an empty form.

src/components/Property/RaiseWorkOrder/RateScheduleItemView.js
```diff
@@ -27,6 +28,7 @@ const RateScheduleItemView = ({
     []
   )
   const [rateScheduleItemCosts, setRateScheduleItemCosts] = useState([])
+  const ADDING_MULTIPLE_SOR_PAGE = 2
```
NeilvB, line 31: Assigning to the same const in multiple places as we have here and in the parent `RaiseWorkOrderFormView.js` can lead us to tricky bugs later on. One const value could be renamed or reassigned but without remembering to update the other. Suggest we instead pass down a function which manages the page transition in reference to the const and give it a specific name like `changePageToAddMultipleSORs` or similar. This also helps readability when reading the parent - it's clear that the child can change to a specific page, but not any page.

src/components/WorkOrder/RateScheduleItems/AddedRateScheduleItems.js
```diff
@@ -11,9 +11,11 @@ const AddedRateScheduleItems = ({
   sorSearchRequest,
   sorCodeArrays,
   setSorCodeArrays,
+  setCurrentPage,
 }) => {
   const [rateScheduleItems, setRateScheduleItems] = useState([...addedTasks])
   const [nextFreeIndex, setNextFreeIndex] = useState(addedTasks.length)
+  const ADDING_MULTIPLE_SOR_PAGE = 2
```
NeilvB, line 18: See other suggestion about this const duplication - better to pass down a specific function prop here which manages the page change instead, without this component knowing about how the parent keeps track of which page to show.

src/components/WorkOrder/RateScheduleItems/AddedRateScheduleItems.js
```diff
@@ -132,6 +138,11 @@ const AddedRateScheduleItems = ({
       <a className="lbh-link" href="#" onClick={addRateScheduleItem}>
         + Add another SOR code
       </a>
+      <br />
+      <br />
```
NeilvB, line 142: Do we just want the two links on separate lines here? If so it's best to avoid using `<br />` as that is a specific use case for new lines amongst text. Could just put both links within their own `<div>`

src/components/Property/RaiseWorkOrder/RateScheduleItemView.js
```diff
@@ -212,6 +219,13 @@ const RateScheduleItemView = ({
           <a className="lbh-link" href="#" onClick={addRateScheduleItem}>
             + Add another SOR code
           </a>
+          <br />
+          <br />
```
NeilvB, line 223: See other comment about `<br />`

cypress/integration/work-order/create/create-with-adding-multiple-sors.spec.js
```diff
@@ -0,0 +1,92 @@
...
+        cy.get('#contractor').type('Purdy Contracts (P) Ltd - PCL')
+        cy.contains('+ Add multiple SOR codes').click()
+      })
+
+      cy.get('#repair-request-form').should('not.exist')
+
+      cy.contains('Add multiple SOR codes')
+      cy.get('#adding-multiple-sors-form').within(() => {
+        cy.contains('Enter SOR codes as a list:')
+      })
```
NeilvB, line 89: I didn't think of it before but please add an assertion within the same test for presence validation of SOR codes which submits the form and expects the validation error message.

cypress/integration/work-order/update-with-adding-multiple-sors.spec.js
```diff
@@ -0,0 +1,54 @@
...
+      cy.get('#repair-request-form').within(() => {
+        cy.contains('+ Add multiple SOR codes').click()
+      })
+
+      cy.get('#repair-request-form').should('not.exist')
+
+      cy.contains('Add multiple SOR codes')
+      cy.get('#adding-multiple-sors-form').within(() => {
+        cy.contains('Enter SOR codes as a list:')
+      })
```
NeilvB, line 51: I didn't think of it before but please add an assertion within the same test for presence validation of SOR codes which submits the form and expects the validation error message.

NeilvB (changes requested):
There's a few requests here but they're all relatively small. I like the new component. new tests and the approach of assigning to the consts to control page views.

As agreed, we won't actually merge this until the validation / deduplication functionality is added. We also need to add a feature flag using the new configuration API - please hold off on that too as we need to confirm how access is going to work for that.

PR #669, opened 2022-05-03
Allow operatives see who they worked with on w/o

siapankina:
Screenshot 2022-05-03 at 22 27 48
ticket: https://www.pivotaltracker.com/story/show/181898163

### DIFF
--- a/src/components/WorkOrder/MobileWorkingWorkOrder.js
+++ b/src/components/WorkOrder/MobileWorkingWorkOrder.js
@@ -106,21 +106,12 @@ const MobileWorkingWorkOrder = ({
               <h4 className="lbh-heading-h4">Overtime work order</h4>
             )}
             {operativesCount > 1 && (
-              <>
-                <h4 className="lbh-heading-h4">Operatives</h4>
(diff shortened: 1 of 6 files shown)


src/components/WorkOrder/MobileWorkingWorkOrder.js
```diff
@@ -104,8 +105,28 @@ const MobileWorkingWorkOrder = ({
...
+                      : 'N/A'
+                    const operativeDisplay = [
+                      operative.name,
+                      percentageDisplay,
+                    ].join(' - ')
+                    return <li className="govuk-list">{operativeDisplay}</li>
+                  })}
+                </ol>
+              </>
+            )}
```
NeilvB, line 124: Can we use the existing `<OperativeList>` component here? It accepts a `readOnly` prop which lists operatives without links and seems very close to what we want, with the exception of some margins. If the new Figma design and the existing non-readonly list differ, it's probably worth having a call with the design team to get consistency between the two kinds of lists - then we can use the same component.

NeilvB (approved):

PR #678, opened 2022-05-06
Prepare raise form for Multiple SORs workflow

NeilvB:
We need this form to retain state when the user navigates away to another form step and comes back, as they will for the upcoming multiple SOR workflow.

This PR prepares for that by pushing state up into the parent "view" component.

It also removes some unnecessary hidden fields, simplifies the way we control input disabled states, and fixes lots of error output in Jest test runs.

### DIFF
--- a/src/pages/api/[...path].test.js
+++ b/src/pages/api/[...path].test.js
@@ -290,7 +290,7 @@ describe('/api/[...path]', () => {
             axios.create = jest.fn(() => axios)
 
             // Mock the 'GET' return value from the service API
-            axios.mockImplementationOnce(() =>
+            axios.mockImplementation(() =>
               Promise.resolve({
                 status: 200,
                 data: { key: 'hackney' },
--- a/reporter-rules.json
+++ b/reporter-rules.json
@@ -14,5 +14,13 @@
   {
     "match": "Component undefined has no permittedRoles defined",
     "group": null
+  },
+  {
+    "match": "Error fetching toggles from configuration API: {\"response\":{\"status\":500}}",
+    "group": null
+  },
+  {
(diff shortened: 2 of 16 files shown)


src/pages/api/[...path].test.js
```diff
@@ -266,17 +266,17 @@ describe('/api/[...path]', () => {
...
+        // 'api/operatives/12345',
+        // 'api/operatives/12345/workorders',
+        // 'api/workOrders',
+        // 'api/workOrders/01234567',
+        // 'api/workOrders/01234567/variation-tasks',
+        // 'api/workOrders/01234567/tasks',
+        // 'api/workOrders/01234567/notes',
       ].forEach((url) => {
-        describe(`requesting the ${url} endpoint`, () => {
+        describe.only(`requesting the ${url} endpoint`, () => {
```
Comment (siapankina) [inline src/pages/api/[...path].test.js:279]: this can be removed :)

src/pages/api/[...path].test.js
```diff
@@ -266,17 +266,17 @@ describe('/api/[...path]', () => {
...
-        'api/workOrders/01234567/notes',
+        // 'api/appointments',
+        // 'api/operatives',
+        // 'api/operatives/12345',
+        // 'api/operatives/12345/workorders',
+        // 'api/workOrders',
+        // 'api/workOrders/01234567',
+        // 'api/workOrders/01234567/variation-tasks',
+        // 'api/workOrders/01234567/tasks',
+        // 'api/workOrders/01234567/notes',
```
Comment (siapankina) [inline src/pages/api/[...path].test.js:277]: Probably this needs to be uncommented?

siapankina (changes requested):
Looks great! 
Happy to approve after comments are addressed

NeilvB:
Thanks @siapankina - I've removed those comments and spec restriction now. Please can you approve?

siapankina (approved):

PR #708, opened 2022-07-26
Add hotjar to RepairsHub FE

Asura5:


LBHCallumM (approved):

src/components/Meta/index.js
```diff
@@ -1,7 +1,12 @@
 import Head from 'next/head'
+import { hotjar } from 'react-hotjar'
+import { useEffect } from 'react'
 
 const Meta = ({ title }) => {
   const applicationTitle = 'Hackney Repairs Hub'
+  useEffect(() => {
+    hotjar.initialize(783901, 6)
+  }, [])
```
pixeltrix, line 9: Surprised this didn't trigger a linter warning - normally it says something like:

```
Warning: React Hook useEffect has a missing dependency: 'hotjar.initialize'. Either include it or remove the dependency array.  react-hooks/exhaustive-deps
```

Maybe there's some difference in config between RHFE and BCFE 🤔

pixeltrix (approved):

src/components/Meta/index.js
Asura5: If I had to guess it would be that BCFE has an extra linter extension in .eslintrc - "`next/core-web-vitals`" which RHFE does not have

PR #714, opened 2022-08-11
Update permissions for contractor

LBHCallumM:
### Description of change

Allow Contractors (Purdy) to search for properties.

Previously, contractors couldn't use the property search, or view any property. The only option was to search with the workOrderReference. I have updated the permissions to:
1. Allow contractors to search for a property
2. Allow contractors to view the property page

The listWorkOrders endpoint already implemented contractor restrictions, so a contractor can only see workOrders that they own.

#### Other changes

In user permissions, I have created a new method `canSearchForProperty`. The method is identical to `canAccessWorkOrder`. I found the use of the method `canAccessWorkOrder` within the search confusing, since accessing workOrders is irrelevant to search. I have created a new method with a more accurate name to resolve this.


cypress/integration/search/search.spec.js
```diff
@@ -260,116 +260,4 @@ describe('Search', () => {
       })
     }
   )
-
-  context('When the role of user allows searching by work order only', () => {
```
Asura5, line 264: Why are all these tests being removed?

LBHCallumM: I have re-added the tests, except for the test that 'shows an error when property search data is entered'

Asura5 (approved):

PR #721, opened 2022-08-26
MR-398 Multiple SORs

Asura5:


src/components/Property/RaiseWorkOrder/BudgetCodeItemView.js
```diff
@@ -74,7 +74,8 @@ const BudgetCodeItemView = ({
 }
 
 BudgetCodeItemView.propTypes = {
-  budgetCodeId: PropTypes.string.isRequired,
+  budgetCodeId: PropTypes.oneOfType([PropTypes.string, PropTypes.number])
+    .isRequired,
```
LBHCallumM, line 78: Is there a chance that this could cause side effects? Surely its better to better to have either string or number.

src/components/Property/RaiseWorkOrder/RateScheduleItemView.js
```diff
@@ -18,12 +18,24 @@ const RateScheduleItemView = ({
...
+    if (arrayOfRateScheduleItemComponentIndexes.length > 1) {
+      if (
+        formState.rateScheduleItems.length <
+        arrayOfRateScheduleItemComponentIndexes.length
+      ) {
+        arrayOfRateScheduleItemComponentIndexes.pop()
+      }
+    }
+  }, arrayOfRateScheduleItemComponentIndexes)
+
```
LBHCallumM, line 38: This could be much cleaner if you negated the if statements. For example:

```suggestion
  useEffect(() => {
    if (arrayOfRateScheduleItemComponentIndexes.length === 0) return;
    if (formState.rateScheduleItems.length === arrayOfRateScheduleItemComponentIndexes.length) return;
      
    arrayOfRateScheduleItemComponentIndexes.pop();
  }, arrayOfRateScheduleItemComponentIndexes)

```

LBHCallumM (approved):

src/components/Property/RaiseWorkOrder/BudgetCodeItemView.js
ange47rm: I set this because there were warnings about the prop type being passed in. 
What you say makes sense, however what was happening was:
- if budgetCodeId was set to PropTypes.string.isRequired - the console Warning would complaining about the component being passed a number (instead of the expected string.
- I switched the PropTypes.string.isRequired to PropTypes.number.isRequired and then the warning was complaining about the value going it being a string and not a number.
Basically it'd complain either way (even if you'd expect the passed in value to be always the same type)... That's why to remove the warning I set it that way.
I agree it kind of goes against what we're trying to do here, defining types as we're making it more vague.

LBHCallumM: Is it complaining because in some places the prop is a string, and other places it is a number? Surely a better solution would be to convert all of the numbers to strings

src/components/Property/RaiseWorkOrder/RateScheduleItemView.js
ange47rm: To my eyes, I think the way Dave wrote it is easier to read but that's subjective

Asura5: Also, the logic is different. In the top example if arrayOfRateScheduleItemComponentIndexes.length > formState.rateScheduleItems.length, the pop won't happen, whereas in the reworked example it will

PR #733, opened 2023-01-10
U2NhZmZvbGQgQmFja09mZmljZSBGcm9udGVuZWQ=

LBHCallumM:
ToDo - change permitted roles once applied


src/pages/backoffice/sor-contracts.js
```diff
@@ -0,0 +1,32 @@
...
+    </>
+  )
+}
+
+export const getServerSideProps = getQueryProps
+
+SORContractsPage.permittedRoles = [
+  AGENT_ROLE,
+  CONTRACTOR_ROLE,
+  CONTRACT_MANAGER_ROLE,
```
Asura5, line 27: Could these be extracted to a shared file? As we're probably going to have the same access level for all parts of the BO

Asura5 (approved):

src/pages/backoffice/sor-contracts.js
Asura5: the roles. just seeing a lot of repetition

PR #735, opened 2023-01-12
Add DataAdmin role to frontend

LBHCallumM:


Asura5 (approved):
is fishery a new library? If so what's the packed size?

LBHCallumM:
> is fishery a new library? If so what's the packed size?

I've essentially just copied the setup for other roles, so 'fishery' was already in use.

PR #739, opened 2023-01-17
Add SorCode form

LBHCallumM:
### Description of change

image

Generate and import a CSV file with the specified headers.

The input isn't overly validated because I didn't want to over-engineer it. We can decide on any further changes once it has been used.


ange47rm (approved):
Nice one, approved

PR #742, opened 2023-02-08
MR-681 - Add multiple SOR codes form v2

ange47rm:
Enable user to add one or multiple new SOR codes

- Added SOR code specific fields (SOR code, cost, SMV, short description, long description) and styling (CSS grid)
- State for each new SOR code on the screen is managed using React's `useReducer`
- Validation for these fields (at the moment, validation just checks that all fields are filled / are not empty - but this can be improved if there are specific validation requirements)
- Added tests
- CSV functional code has been commented out (JSX code removed), but this can be added back in if required. Rather than leave this on the same page, we could possibly have a radio button to let the user choose whether to use the CSV upload functionality or do it through the form (and switching 'mode' would clear the components' state).

Normal view (one blank new SOR code is shown by default on page load)
image

Validation:
image


src/components/BackOffice/AddSORCodes/index.js
```diff
@@ -90,50 +122,172 @@ const AddSORCodes = () => {
...
+    // Used for CSV bulk upload
+    // if (loading || fileLoading) return
 
-    var errors = validate()
-    setErrors(errors)
+    const formValid = validateForm()
 
-    // there must be no errors
-    if (Object.keys(errors).length > 0) return
+    if (formValid == false) return
```
LBHCallumM, line 275: You could simplify this with `if (!validateForm())`

ange47rm: Damn I've only just seen this (post merge), but good spot

PR #743, opened 2023-02-14
Updates to Property Contracts page [BackOffice]

LBHCallumM:
[Ticket](https://hackney.atlassian.net/browse/MR-680?atlOrigin=eyJpIjoiMjQxMDdiN2Q2MzM3NGQ1Yjg5NWZjZTg0YTFkMTNhZjIiLCJwIjoiaiJ9)

### Description of change

- Moved reused requests to `requests.js` file
- Created `useSelectContract` hook that is used in multiple locations
- Refactored SorContracts form


src/components/BackOffice/SORContracts/utils.test.js
```diff
@@ -0,0 +1,120 @@
...
+      // Act
+      const response = validatePropertyReference(propertyReference)
+
+      // Assert
+      expect(response).toBe(false)
+    })
+  })
+
+  describe('copyContractsSelected', () => {
+    it('returns true', () => {
```
ange47rm, line 35: No biggie but I would have probably added a little more other than just 'return true/false' to give a little bit more context on what's being tested (I know it's the function it utils.js). Same applies for lines 45 57 67 74

cypress/integration/backoffice/sor-contracts.spec.js
```diff
@@ -0,0 +1,186 @@
...
+
+    const propertyReference1 = '12345678'
+    const propertyReference2 = '87654321'
+    cy.get('[data-test="sourcePropertyReference"]').type(propertyReference1)
+    cy.get('[data-test="destinationPropertyReference"]').type(
+      propertyReference2
+    )
+    cy.get("[data-test='submit-button']").click()
+
+    cy.wait('@sorContractRequest')
```
ange47rm, line 91: There are no assertions in this test. If the purpose of the test is to assert that it sends a request I'd set up an assertion here such as 

cy.wait('@sorContractRequest')
      .its('request.path')
      .should('deep.equal', '/api/backOffice/sor-contracts')

cypress/integration/backoffice/sor-contracts.spec.js
```diff
@@ -0,0 +1,186 @@
...
+
+    cy.contains('SOR Contract Modification')
+    cy.contains('SOR contracts modified successfully!')
```
ange47rm, line 94: In my understanding the cy.contains itself is not an assertion, but it's used to get HTML elements. If you want to assert that the text "SOR Contract Modification" is shown I'd suggest adding a "should" after it:

example: cy.contains('SOR Contract Modification').should('be.visible')

or you could do something like 

example: cy.get('[data-test="blabla"]').should('have.text', 'SOR Contract Modification')

I'd apply this to every cy.contains in this file that's been used as an assertion

ange47rm (approved):
Looks good, nice job with the selectContract hook. I'll approve this however I strongly suggest to set up the assertions correctly in 'sor-contracts.spec.js' as per comment

PR #744, opened 2023-02-15
MR-688 - Bulk close work order confirmation modal

ange47rm:
Implemented confirmation modal when a user submits a request to close work orders

image

### DIFF
--- a/src/components/BackOffice/CloseWorkOrders/index.js
+++ b/src/components/BackOffice/CloseWorkOrders/index.js
@@ -95,22 +95,27 @@ const CloseWorkOrders = () => {
           title="Are you sure?"
           isOpen={showDialog}
           onDismiss={() => setShowDialog(false)}
+          data-test="confirmation-modal"
         >
           <h3 className="lbh-heading-h3">
-            Are you sure you want to change the status of the following Work
-            Orders to "
+            Are you sure you want to change the status of these Work Orders to "
             {selectedOption == 'CloseToBase' ? 'Completed' : 'Cancelled'}"?
           </h3>
-          <p className="lbh-body">{workOrderReferences}</p>
 
           <div className="lbh-dialog__actions">
-            <a href="#" className="govuk-button lbh-button" onClick={submit}>
+            <a
+              href="#"
+              className="govuk-button lbh-button"
+              onClick={submit}
+              data-test="confirm-button"
+            >
               Confirm
             </a>
 
             <button
               onClick={() => setShowDialog(false)}
               className="lbh-link lbh-link--no-visited-state"
+              data-test="cancel-button"
             >
               Cancel
             </button>
--- a/src/components/BackOffice/CloseWorkOrders/index.js
+++ b/src/components/BackOffice/CloseWorkOrders/index.js
@@ -98,7 +98,7 @@ const CloseWorkOrders = () => {
           data-test="confirmation-modal"
         >
           <h3 className="lbh-heading-h3">
-            Are you sure you want to change the status of these Work Orders to "
+            The status of these Work Orders will change to "
             {selectedOption == 'CloseToBase' ? 'Completed' : 'Cancelled'}"?
           </h3>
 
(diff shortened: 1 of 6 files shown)


src/components/BackOffice/CloseWorkOrders/index.js
```diff
@@ -88,7 +88,39 @@ const CloseWorkOrders = () => {
     return newErrors
   }
 
-  const handleSubmit = (event) => {
+  const renderConfirmationModal = () => {
```
Asura5, line 91: Remember there are instances where 300+ work orders are being closed at the same time

ange47rm: Thanks, might be worth removing the list of work orders!

src/components/BackOffice/CloseWorkOrders/index.js
```diff
@@ -132,36 +173,39 @@ const CloseWorkOrders = () => {
...
                   onClick={() => setFormSuccess(null)}
                 >
-                  Bulk-close more workOrders
+                  Bulk-close more Work Orders
                 </a>
               </p>
             </div>
           ) : (
-            <form onSubmit={handleSubmit}>
+            <form onSubmit={validateForm}>
```
LBHCallumM, line 201: I don't think validateForm entirely explains what this method does. HandleSubmit is much better

src/components/BackOffice/CloseWorkOrders/index.js
```diff
@@ -88,7 +88,44 @@ const CloseWorkOrders = () => {
...
+        <Dialog
+          title="Are you sure?"
+          isOpen={showDialog}
+          onDismiss={() => setShowDialog(false)}
+          data-test="confirmation-modal"
+        >
+          <h3 className="lbh-heading-h3">
+            Are you sure you want to change the status of these Work Orders to "
+            {selectedOption == 'CloseToBase' ? 'Completed' : 'Cancelled'}"?
+          </h3>
```
LBHCallumM, line 103: The wording needs to be more direct. I don't think this is enough to scare people into being 100% sure they're doing the right thing.

I suggest something more like "Permanently Cancel WorkOrders", or "Complete WorkOrders". Also, the confirmation button should say something like 'Delete workOrders'. Confirm is too vague.

This article is pretty helpful:
https://uxdesign.cc/are-you-sure-you-want-to-do-this-microcopy-for-confirmation-dialogues-1d94a0f73ac6

LBHCallumM: Also, could you add this for the two other forms

src/components/BackOffice/CloseWorkOrders/index.js
ange47rm: I needed to link the "Yes button" of the modal to an action, so I've put validateForm first, to validate the form for errors, and if there are no errors, then the modal is shown. At this point you know the form has been populated correctly and there are no errors. If then the user presses "Yes/Confirm" the network request is fired. I think it's ok as is but if you've got any suggestions on how to name this method let me know your suggestions

src/components/BackOffice/CloseWorkOrders/index.js
```diff
@@ -88,7 +88,24 @@ const CloseWorkOrders = () => {
...
 
-  const handleSubmit = (event) => {
+  const renderConfirmationModal = () => {
+    if (showDialog) {
+      return (
+        <ConfirmationModal
+          title={'Permanently close Work Orders?'}
+          showDialog
+          setShowDialog={setShowDialog}
+          modalText={`The status of the selected Work Orders will change to "${
```
Asura5, line 98: Work Orders is capitalised throughout - is there a reason for this? It's not a proper noun.

ange47rm: I agree it shouldn't be, but I've capitalised it to keep it consistent with the rest of the solution. Happy to change it everywhere to "work orders"

Asura5: yeah change it to no caps

ange47rm: Will do

ange47rm: Done

Asura5 (approved):

PR #747, opened 2023-02-16
MR-688 - Modify SOR contracts confirmation modal

ange47rm:
- Added confirmation modal prior to sending POST request to modify SOR contracts
- Added tests to assert the modal is shown and other test assertions mentioned in PR comments #743 
- Corrected issue with useSelectContract hook to prevent error when resetForm() would call handleSelectContractor and pass null instead of a valid browser event.

image

### DIFF
--- a/src/components/BackOffice/hooks/useSelectContract.js
+++ b/src/components/BackOffice/hooks/useSelectContract.js
@@ -46,7 +46,6 @@ const useSelectContract = () => {
   }, [selectedContractor])
 
   const handleSelectContractor = (e) => {
-    // Ternary operator prevents the line below from erroring when "resetForm() calls handleSelectContractor(null)"
     const contractorName = e ? e.target?.value : undefined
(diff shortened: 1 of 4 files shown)


src/components/BackOffice/hooks/useSelectContract.js
```diff
@@ -46,7 +46,8 @@ const useSelectContract = () => {
   }, [selectedContractor])
 
   const handleSelectContractor = (e) => {
-    const contractorName = e.target.value
+    // Ternary operator prevents the line below from erroring when "resetForm() calls handleSelectContractor(null)"
+    const contractorName = e ? e.target?.value : undefined
```
ange47rm, line 50: Ternary operator prevented the line below from erroring when "resetForm() calls handleSelectContractor(null)"
After this change the form is reset correctly (manually tested)

Asura5 (approved):

LBHCallumM (approved):

PR #749, opened 2023-02-17
RH - Improvements to Backoffice testing (specific to form reset)

ange47rm:
The contract was not being reset correctly after a successful "Modify SOR Contracts" workflow. There were no tests to test the state of the form after a successful submit, and a reset.

- Added handleFormReset method to useSelectContract hook to set Contractor and Contract values back to null after form reset. 
- Added "reset form" test for Add new SOR codes
- Added test for reset form for the Modify SOR contracts feature. Implemented use of handleFormReset in component.
- Spelling/casing minor corrections
- Added a few assertions where only the `cy.contains` was being used (as an assertion). Cy.contains is used to return an HTML element, assertions should be followed by a "should".

### DIFF
--- a/src/components/BackOffice/hooks/useSelectContract.js
+++ b/src/components/BackOffice/hooks/useSelectContract.js
@@ -46,7 +46,7 @@ const useSelectContract = () => {
   }, [selectedContractor])
 
   const handleSelectContractor = (e) => {
-    const contractorName = e ? e.target?.value : undefined
+    const contractorName = e.target.value
     const selectedContractor = contractors.find(
       (contractor) => contractor.contractorName === contractorName
     )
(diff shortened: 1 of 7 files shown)


src/components/BackOffice/hooks/useSelectContract.js
```diff
@@ -46,7 +46,7 @@ const useSelectContract = () => {
   }, [selectedContractor])
 
   const handleSelectContractor = (e) => {
-    const contractorName = e ? e.target?.value : undefined
+    const contractorName = e.target.value
```
ange47rm, line 49: This was changed in PR #747 , now being set back to it as handleSelectContractor is no longer called when the reset form happens. Instead we call the new function handleFormReset which has the only task to reset the 2 values (contractor, contract)

Asura5 (approved):

PR #752, opened 2023-04-21
MR-697 - Add StartTime to CloseWorkOrder form

LBHCallumM:
## Summary of Changes

- Fix snapshot date issue (sets the timezone to UTC when `NODE_ENV=test`
- Add StartTime to the main workOrder view (above appointments)
- Add StartDate and StartTime fields to CloseWorkOrderComponent (StartDate optional, StartTime required when StartDate not null)
- Update SummaryCloseWorkOrder component to include StartTime
- Sends a request to `/api/workOrders/starttime` when a workOrder is closed
- Refactored `ControlledTimeInput` component
- Added dayjs library (better alternative to moments), for use in the Summary page
- Update Cypress tests

### WorkOrder page
image

### Close workOrder form
image

## Summary page
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "axios": "^0.21.1",
     "cookie": "^0.4.1",
     "date-fns": "^2.16.1",
+    "dayjs": "^1.11.7",
     "dotenv": "^8.2.0",
     "easy-soap-request": "^4.1.3",
     "http-status-codes": "^2.1.4",
--- a/src/components/Form/TimeInput/index.js
+++ b/src/components/Form/TimeInput/index.js
@@ -23,6 +23,7 @@ const TimeInput = forwardRef(
       register,
       onChange,
       required,
+      showAsOptional = false,
       ...otherProps
     },
     ref
@@ -53,7 +54,8 @@ const TimeInput = forwardRef(
           <legend
             className={`govuk-fieldset__legend govuk-fieldset__legend--${labelSize}`}
           >
-            {label} {required && <span className="govuk-required">*</span>}
+            {label} {showAsOptional && '(optional)'}{' '}
+            {required && <span className="govuk-required">*</span>}
           </legend>
           <span id={`${name}-hint`} className="govuk-hint lbh-hint">
             {hint}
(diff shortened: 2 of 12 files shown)


package.json
```diff
@@ -31,6 +31,7 @@
     "axios": "^0.21.1",
     "cookie": "^0.4.1",
     "date-fns": "^2.16.1",
+    "dayjs": "^1.11.7",
```
Asura5, line 34: What method did you use to make this conclusion? On what basis is this better than moment? On the basis that everyone is familiar with moment and non of the devs have used dayJS before?

LBHCallumM: Firstly, according to the[ readme](https://yarnpkg.com/package/moment), `Moment.js` is now a legacy project. They recommend using a different library.

I selected `dayjs` because it was built to use the same API at Moment. So I see no issue for other developers to use it.

Asura5: Good logic

src/components/Form/TimeInput/index.js
```diff
@@ -135,20 +135,35 @@ TimeInput.propTypes = {
...
+            // not required - valid if empty
+            if (!required && !value) return true
+
+            const hourStr = value.split(':')[0] || ''
+            const minuteStr = value.split(':')[1] || ''
+
+            const hour = parseInt(hourStr)
+            const minute = parseInt(minuteStr)
+
+            if (isNaN(hour) || isNaN(minute)) return 'Please enter a valid time'
```
Asura5, line 165: String is duplicated with line 178. Also, if emptyErrorMessage is a prop, why isn't the invalidErrorMessage? Happy with either approach, but we are either promoting this component for re-useability or we are not

LBHCallumM: I was trying to use the label in the error message but had issues with '(optional)' being part of the label. 

I have replaced the first error with the following:
```javascript
if (required && !value) return `Please enter a ${label.toLowerCase()}`
```

Then added a `showAsOptional` parameter to the component.

src/components/WorkOrders/CloseWorkOrderByProxy.js
```diff
@@ -63,6 +66,17 @@ const CloseWorkOrderByProxy = ({ reference }) => {
         })
       }
 
+      if (startDate !== '') {
+        await frontEndApiRequest({
+          method: 'put',
+          path: `/api/workOrders/starttime`,
+          requestData: {
+            startTime: startDate,
```
ange47rm, line 74: startDate in this case includes the time too correct? If yes I'd consider renaming this to startDateAndTime

LBHCallumM: scroll down

src/components/WorkOrders/SummaryCloseWorkOrder.test.js
```diff
@@ -35,8 +35,8 @@ describe('SummaryCloseWorkOrder component', () => {
       <SummaryCloseWorkOrder
         reference={props.reference}
         notes={props.notes}
-        time={props.time}
-        date={props.date}
+        completionDate={props.completionDate}
```
ange47rm, line 38: I know it's no new addition but I feel the title of this test makes me think there should be an assertion to verify these rows don't exist but there are no assertion other than a snapshot comparison to what's rendered

ange47rm (approved):
Approving with comments

PR #760, opened 2023-07-13
Show boiler house flag

LBHCallumM:
## Ticket

[https://hackney.atlassian.net/browse/MR-716?atlOrigin=eyJpIjoiZTU0YTcwYzI0OTcyNDEyOGJlZGM3YjMyM2Q1NGZmN2UiLCJwIjoiaiJ9](https://hackney.atlassian.net/browse/MR-716?atlOrigin=eyJpIjoiZTU0YTcwYzI0OTcyNDEyOGJlZGM3YjMyM2Q1NGZmN2UiLCJwIjoiaiJ9)

## Summary of Changes
- Add boilerHouseFlag to PropertyDetails component
- The TenureDetails component was renamed to PropertyFlag. The new name more accurately describes the component
- Refactored PropertyDetailsGrid to fill full width on mobile

### Property with boilerHouse flag (desktop)

image

### Property with boilerHouse flag (mobile)

image

#### Before

image

#### After

image

### DIFF
--- a/src/components/Property/PropertyFlags/index.js
+++ b/src/components/Property/PropertyFlags/index.js
@@ -1,12 +1,12 @@
 import PropTypes from 'prop-types'
 import { frontEndApiRequest } from '@/utils/frontEndApiClient/requests'
-import Alerts from './Alerts'
-import Tenure from './Tenure'
-import TenureDetail from './TenureDetail'
+import Alerts from '../Alerts'
+import Tenure from '../Tenure'
+import TenureDetail from '../TenureDetail'
 import { useEffect, useState } from 'react'
 import Spinner from '@/components/Spinner'
 import ErrorMessage from '@/components/Errors/ErrorMessage'
-import PropertyBoilerHouseDetails from './PropertyBoilerHouseDetails'
+import PropertyBoilerHouseDetails from '../PropertyBoilerHouseDetails'
 
 const PropertyFlags = ({
   canRaiseRepair,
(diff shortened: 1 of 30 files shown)


src/components/Property/PropertyFlags/index.js
```diff
@@ -1,17 +1,19 @@
...
-import TenureDetail from './TenureDetail'
+import Alerts from '../Alerts'
+import Tenure from '../Tenure'
+import TenureDetail from '../TenureDetail'
 import { useEffect, useState } from 'react'
 import Spinner from '@/components/Spinner'
 import ErrorMessage from '@/components/Errors/ErrorMessage'
+import PropertyBoilerHouseDetails from '../PropertyBoilerHouseDetails'
 
-const TenureDetails = ({
```
Asura5, line 10: It might just be worth surfacing this [good] renaming decision a bit more widely, as several teams have worked with this in the past and might be confused

src/components/Property/PropertyFlags/index.js
```diff
@@ -76,6 +78,8 @@ const TenureDetails = ({
       <Alerts alerts={personAlerts} alertType="Contact" />
     )
 
+  const showBoilerHouseDetails = () => boilerHouseId !== ''
```
Asura5, line 81: Would it be worth including a check that the asset type is boilerhousable, as well as the ID being not null?

Asura5 (approved):

src/components/Property/PropertyFlags/index.js
LBHCallumM: I could, however, I don't think this is necessary. There is already a restriction on MMH when adding a boiler house. to a property

src/components/Property/PropertyFlags/index.js
LBHCallumM: Who do you think this would be relevant for? Team houseboat?

PR #770, opened 2023-09-04
MR-740 Feat: One Job At a Time - Frontend

Asura5:
- Implement one job at a time
- The user's profile (returned by /hub-user) now contains a boolean IsOneJobAtATime flag, which is used to selectively show/hide the job list

### DIFF
--- a/cypress/integration/home-page.spec.js
+++ b/cypress/integration/home-page.spec.js
@@ -549,8 +549,7 @@ describe('Home page', () => {
           cy.get('li').eq(0).click()
         })
 
-        cy.wait(['@operativesWorkOrder'])
-        cy.contains('WO 10000625')
+        cy.contains('10000625')
       })
     })
 
--- a/cypress/integration/home-page.spec.js
+++ b/cypress/integration/home-page.spec.js
@@ -546,11 +546,11 @@ describe('Home page', () => {
               cy.contains('Lorem ipsum dolor sit amet, consectetur efficitur.')
             })
 
-          cy.get('li').eq(1).click()
+          cy.get('li').eq(0).click()
         })
 
         cy.wait(['@operativesWorkOrder', '@property', '@task'])
-        cy.contains('WO 10000626')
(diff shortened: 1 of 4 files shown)


cypress/integration/home-page.spec.js
```diff
@@ -452,4 +452,132 @@ describe('Home page', () => {
...
+            fixture: 'workOrders/task.json',
+          }
+        ).as('task')
+      })
+
+      it('Displays only one work order appointment', () => {
+        cy.visit('/')
+        cy.wait('@operativesWorkOrders')
+
+        cy.get('.lbh-heading-h2').contains('Friday 11 June')
```
ange47rm, line 525: In a few occasions in this test file "contains" seems to be used to make an assertion, but "contains" itself, to my understanding, is just an easier way to get an item, like "cy.get", using the item's text. It doesn't count as an assertion though.

I would replace all the "contains assertions" with actual assertions using the "should" keyword.

For example:
- cy.contains('Friday 11 June').should('exist')
or
- cy.get('.lbh-heading-h2').should('have.text', 'Friday 11 June')

The test will fail regardless if the text can't be found by the get/contains, so it's not a big deal

Asura5:
> Looks good, let me know once the failing test has been corrected. It seems like the test still finds 3 items when only 1 is expected

Ah yeah, it's counting the completed / no access items - will fix that

cypress/integration/home-page.spec.js
Asura5: I assume you would advocate for replacing this across every spec file then, in order to keep the files consistent?

ange47rm: If the same "contains assertions" have been used elsewhere yes I would. If you think it's multiple files involved then maybe it's worth adding it as a tech debt item

Asura5: agreed

ange47rm (approved):

PR #772, opened 2023-09-18
Pass `withContactDetails` query on the raise workOrder page

LBHCallumM:
### Description of change

Pass `withContactDetails` flag when raising a repair. Backend change requires this query parameter to include any contact details.

### DIFF
--- a/src/components/Property/RaiseWorkOrder/RaiseWorkOrderFormView.js
+++ b/src/components/Property/RaiseWorkOrder/RaiseWorkOrderFormView.js
@@ -136,6 +136,9 @@ const RaiseWorkOrderFormView = ({ propertyReference }) => {
       const data = await frontEndApiRequest({
         method: 'get',
         path: `/api/properties/${propertyReference}`,
+        params: {
+          withContactDetails: true
+        }
       })
       const priorities = await frontEndApiRequest({
         method: 'get',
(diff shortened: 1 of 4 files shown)


src/components/Property/RaiseWorkOrder/RaiseWorkOrderFormView.js
```diff
@@ -136,6 +136,9 @@ const RaiseWorkOrderFormView = ({ propertyReference }) => {
       const data = await frontEndApiRequest({
         method: 'get',
         path: `/api/properties/${propertyReference}`,
+        params: {
+          withContactDetails: true,
+        },
       })
```
ange47rm, line 142: Could there be any other places in RH where the `frontEndApiRequest` needs to be edited or is this the only instance where we do require contact details?

ange47rm (approved):
Looks fine.
Before merging please:
- fix failing test
- make sure there are no other instances where contact details are required when calling the GetProperty endpoint

src/components/Property/RaiseWorkOrder/RaiseWorkOrderFormView.js
LBHCallumM: Nope. Only when raising a repair. The contact details information is provided as a suggestion when filling out the workOrder contact details. After the repair is raised, contact information is stored in repairs hub directly

PR #774, opened 2023-09-25
Fetch contact details from new endpoint

LBHCallumM:
## Summary of changes

Fetch contact details directly from the contact details endpoint when raising a repair.

- Fetch contact details if `tenure.id` is returned in the property response
- Copy existing nextjs authorization to redact contact details if the user is unauthorized + tests

## Additional changes

- Added the four requests into a `Promise.all()` block so that the requests can run simultaneously (instead of in serial)/

## Next steps

Once these changes are deployed, contact details can be removed from the properties endpoint.

### DIFF
--- a/src/components/Property/RaiseWorkOrder/RaiseWorkOrderFormView.js
+++ b/src/components/Property/RaiseWorkOrder/RaiseWorkOrderFormView.js
@@ -159,8 +159,6 @@ const RaiseWorkOrderFormView = ({ propertyReference }) => {
         })
       }
 
-      
-
       setTenure(propertyResponse.tenure)
       setProperty(propertyResponse.property)
       setPriorities(priorities)
--- a/src/components/Property/RaiseWorkOrder/RaiseWorkOrderFormView.js
+++ b/src/components/Property/RaiseWorkOrder/RaiseWorkOrderFormView.js
@@ -133,7 +133,7 @@ const RaiseWorkOrderFormView = ({ propertyReference }) => {
     setError(null)
 
     try {
-      const data = await frontEndApiRequest({
+      const propertyResponse = await frontEndApiRequest({
         method: 'get',
         path: `/api/properties/${propertyReference}`,
       })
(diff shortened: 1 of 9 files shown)


src/components/Property/RaiseWorkOrder/RaiseWorkOrderFormView.js
```diff
@@ -133,34 +133,45 @@ const RaiseWorkOrderFormView = ({ propertyReference }) => {
...
+        frontEndApiRequest({
+          method: 'get',
+          path: `/api/schedule-of-rates/trades?propRef=${propertyReference}`,
+        }),
+        frontEndApiRequest({
+          method: 'get',
+          path: '/api/hub-user',
+        }),
+      ])
+
```
LBHCallumM, line 154: Requests can be chained together

src/components/Property/RaiseWorkOrder/RaiseWorkOrderFormView.js
```diff
@@ -133,34 +133,45 @@ const RaiseWorkOrderFormView = ({ propertyReference }) => {
...
+      let contactDetails = []
+
+      if (propertyResponse?.tenure?.id != null) {
+        contactDetails = await frontEndApiRequest({
+          method: 'get',
+          path: `/api/contact-details/${propertyResponse?.tenure?.id}`,
+        })
+      }
```
LBHCallumM, line 162: Fetch contact details if tenureId is present. This is the same implementation used on the properties endpoint.

src/components/Property/RaiseWorkOrder/RaiseWorkOrderFormView.js
```diff
@@ -133,34 +133,45 @@ const RaiseWorkOrderFormView = ({ propertyReference }) => {
...
 
-      setTenure(data.tenure)
-      setProperty(data.property)
+      setTenure(propertyResponse.tenure)
+      setProperty(propertyResponse.property)
       setPriorities(priorities)
       setTrades(trades)
-      setContacts(data.contactDetails)
+      setContacts(contactDetails)
```
Asura5, line 168: Given your setting of contactDetails is conditional, contactDetails could be null here. Does setContact account for that, or should this also be wrapped with a not null?

src/pages/api/contact-details/[id]/index.test.js
```diff
@@ -0,0 +1,171 @@
...
+    const signedCookie = jsonwebtoken.sign(
+      {
+        name: 'name',
+        email: 'name@example.com',
+        groups: [CONTRACTORS_GOOGLE_GROUPNAME],
+      },
+      HACKNEY_JWT_SECRET
+    )
+
+    test('response contact information is redacted', async () => {
```
Comment (Asura5) [inline src/pages/api/contact-details/[id]/index.test.js:149]: Is this testing the contact details endpoint, or the auth mechanisms & security?

Asura5 (approved):

src/components/Property/RaiseWorkOrder/RaiseWorkOrderFormView.js
LBHCallumM: The response class used by the API is unchanged, so it should behave identically. 

I have double checked this The API will return an empty array if no contact details are found

PR #776, opened 2023-10-09
Adjust by start datetime on FE

Asura5:
- For one job at a time, jobs should be ordered by assigned start and end time
- Recently surfaced from backend under [https://github.com/LBHackney-IT/repairs-api-dotnet/pull/1054](https://github.com/LBHackney-IT/repairs-api-dotnet/pull/1054
)
Context: [https://hackney.atlassian.net/browse/MR-740](https://hackney.atlassian.net/browse/MR-740)
-

### DIFF
--- a/src/components/WorkOrders/MobileWorkingWorkOrdersView.js
+++ b/src/components/WorkOrders/MobileWorkingWorkOrdersView.js
@@ -95,7 +95,12 @@ const MobileWorkingWorkOrdersView = () => {
               <ol className="lbh-list mobile-working-work-order-list">
                 {renderWorkOrderListItems(
                   currentUser.isOneJobAtATime
-                    ? inProgressWorkOrders.slice(0, 1)
+                    ? inProgressWorkOrders.sort((a, b => {
+                      const dateA = new Date(a.assignedStartTime);
+                      const dateB = new Date(b.assignedStartTime);
+                      // Get first job by assigned start time
+                      return dateA - dateB;
+                    }).slice(0, 1))
                     : inProgressWorkOrders
                 )}
                 {renderWorkOrderListItems(visitedWorkOrders)}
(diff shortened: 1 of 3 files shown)


src/components/WorkOrders/MobileWorkingWorkOrdersView.js
```diff
@@ -95,7 +95,12 @@ const MobileWorkingWorkOrdersView = () => {
               <ol className="lbh-list mobile-working-work-order-list">
                 {renderWorkOrderListItems(
                   currentUser.isOneJobAtATime
-                    ? inProgressWorkOrders.slice(0, 1)
+                    ? inProgressWorkOrders.sort((a, b => {
+                      const dateA = new Date(a.assignedStartTime);
+                      const dateB = new Date(b.assignedStartTime);
+                      // Get first job by assigned start time
+                      return dateA - dateB;
+                    }).slice(0, 1))
```
Asura5, line 103: if anyone can think of a more succinct way to to this with strings of dates I'd be interested to hear it

adamtry: Could do `return new Date(a.assignedStartTime) - new Date(b.assignedStartTime);` directly but I think as you have it now is more readable. Could consider pulling it into a function?

adamtry (approved):

LBHCallumM (approved):

src/components/WorkOrders/MobileWorkingWorkOrdersView.js
LBHCallumM: I actually quite like this implementation, verbosity is usually easier to understand. However, it would be better in a function. That way, you get the best of both worlds

Asura5: Refactored in to a new method

PR #782, opened 2023-10-17
Next use case - OJAAT

Asura5:
Frontend change for https://github.com/LBHackney-IT/repairs-api-dotnet/pull/1056 - description is contained there

Logic now is:

If the operative is NOT an OJAAT operative
----THEN display all the work orders

otherwise, if the operative IS an OJAAT operative
----THEN if the operative has started a work order
--------THEN display that work order
----Otherwise, display the next unstarted work order vis-a-vis diary time


src/components/WorkOrders/MobileWorkingWorkOrdersView.js
```diff
@@ -95,13 +102,15 @@ const MobileWorkingWorkOrdersView = () => {
...
+                      : inProgressWorkOrders
+                          .sort((a, b) =>
+                            a.appointment.assignedStart.localeCompare(
+                              b.appointment.assignedStart
+                            )
                           )
-                        )
-                        .slice(0, 1)
+                          .slice(0, 1)
                     : inProgressWorkOrders
```
LBHCallumM, line 114: I think this would be better as a function, with explicit comments. This is going to be very important logic, why not make it as clear to understand as the logic you mentioned above? Perhaps a bit over the top, maybe. But who knows who might be looking at this code.

Suggestion:
```js
  const sortWorkOrderItems = () => {
    // The operative is NOT an OJAAT operative
    if (!currentUser.isOneJobAtATime) return inProgressWorkOrders;

    // If the operative has started a work order
    if (startedWorkOrders?.any()) return startedWorkOrders

    // Return the next unstarted work order 
    return inProgressWorkOrders.sort((a, b) => {
      return a.appointment.assignedStart.localeCompare(
        b.appointment.assignedStart
      )
    })
    .slice(0, 1)  
  }
```

LBHCallumM (approved):
Approved with comment

src/components/WorkOrders/MobileWorkingWorkOrdersView.js
Asura5: Refactored out to a method

PR #784, opened 2023-11-01
Add Remove/Edit functionality to contact details in raise repair workflow

LBHCallumM:
## Ticket

[https://hackney.atlassian.net/browse/MR-761?atlOrigin=eyJpIjoiMGJmOGUwZWRhNzU2NGVjMjlmNzM0MWUxZDlhZjcxMzciLCJwIjoiaiJ9](https://hackney.atlassian.net/browse/MR-761?atlOrigin=eyJpIjoiMGJmOGUwZWRhNzU2NGVjMjlmNzM0MWUxZDlhZjcxMzciLCJwIjoiaiJ9)

## Summary of Changes

Update contacts component and add edit/delete functionality.

- Messaging updated to clearly differentiate between tenant contacts and repair contacts
- The contact details request was removed from the component `RaiseWorkOrderFormView`, and moved down to `Contacts`. 
- Moved `ConfirmationModal` from backoffice, into its own component.
- Separated tenants from household members. Household members are displayed in the existing table. Tenants are displayed with the new view.
- Each person can be edited via a link to MMH. This link will open in a new tab. For this, I have been required to use the `NEXT_PUBLIC_MMH_FRONTEND_URL` environment variable
- To solve the page reload issue, I have implemented the hook `usePageVisibility`. When the page becomes visible again, the contact details will be updated. This workflow works because the user would have needed to visit the other tab, then revisit repairs hub after the change.
- Each phone number can be set as main, or removed. For each, a confirmation modal is shown. Set as main contact changes the subType to `mainNumber`. For this reason, the modal is very specific that it is the contact type that is being updated.
- To forward the PATCH and DELETE requests, I needed to use the `CONTACT_DETAILS_API_URL` environment variable. This approach didn't work with the existing `serviceAPIRequest` method because that hardcoded the URL to go to repairs api. To get around it, I have added `externalAPIRequest`. The new method is functionally the same, except it wont include `REPAIRS_SERVICE_API_URL` in the URL. 

### Contacts view
image

### Set as main modal
image

### Remove contact modal
image

## Local changes
The following environment variables will need to be added to .env.local:

- `NEXT_PUBLIC_MMH_FRONTEND_URL=#`
- `CONTACT_DETAILS_API_URL=#`


src/components/Property/Contacts/TenantsContactTable/modals/SetAsMainModal.js
```diff
@@ -0,0 +1,79 @@
...
+    <ConfirmationModal
+      title={'Set as main contact'}
+      showDialog={showModal}
+      setShowDialog={setShowModal}
+      isLoading={isLoading}
+      modalError={modalError}
+      modalText={
+        <p className="govuk-body">
+          Are you sure you want to change the contact type from{' '}
+          <strong style={{ textTransform: 'capitalize' }}>
```
adamtry, line 66: This may be a bit easier to read with these <strong> subsections set as variables before the return

src/components/Property/Contacts/TenantsContactTable/TenantPhoneNumber.js
```diff
@@ -0,0 +1,69 @@
...
+        </div>
+      )}
+      <div className="tenantContactsTable-buttonGroup">
+        <button
+          type="button"
+          className="tenantContactsTable-button"
+          onClick={() => setShowMainContactModal(true)}
+          disabled={isMainNumber}
+          style={
+            isMainNumber
```
adamtry, line 31: Could this be simplified to `{isMainNumber && {color: '#888'}}`?

src/components/Property/Contacts/TenantsContactTable/TenantsContactTable.test.js
```diff
@@ -0,0 +1,59 @@
...
+import TenantsContactTable from './TenantsContactTable'
+
+describe('TenantsContactTable component', () => {
+  describe('when supplied with an empty tenants list', () => {
+    const tenants = []
+
+    it('renders text indicating that no tenants are available', () => {
+      const { asFragment } = render(<TenantsContactTable tenants={tenants} />)
+
+      expect(asFragment()).toMatchSnapshot()
```
adamtry, line 11: Wouldn't explicit assertions be better here than snapshot comparisons? Or good to have in addition to snapshot comparisons?

src/components/Property/Contacts/TenantsContactTable/modals/SetAsMainModal.js
LBHCallumM: Im not sure what you mean. Could you provide an example?

src/components/Property/Contacts/TenantsContactTable/TenantPhoneNumber.js
LBHCallumM: I think using a class is a much better approach.

image

src/components/Property/Contacts/TenantsContactTable/TenantsContactTable.test.js
LBHCallumM: I think for these static use cases, snapshot tests are better. They are much quicker to implement, and require less maintenance. 

Is there reasons why assertion tests would be better? Is there something im not testing?

src/components/Property/Contacts/TenantsContactTable/modals/SetAsMainModal.js
adamtry: I was thinking of moving them to mini components and just doing:
```js
<p className="govuk-body">
  Are you sure you want to change the contact type from
  <PhoneNumberSubType /> to <MainNumber />?
</p>
```
but the components for this are too large so it doesn't actually add clarity

src/components/Property/Contacts/TenantsContactTable/TenantPhoneNumber.js
adamtry: I saw your commit and it like a good solution, agreed!

src/components/Property/Contacts/TenantsContactTable/TenantsContactTable.test.js
adamtry: True, for these static cases snapshots make sense. I was probably thinking for more interactive cases.

cypress/integration/work-order/create/create.spec.js
```diff
@@ -1759,6 +1766,313 @@ describe('Raise repair form', () => {
...
+        '@sorPrioritiesRequest',
+        '@tradesRequest',
+      ])
+
+      // remove first tenant
+      cy.contains('Remove').first().click()
+
+      // assert modal open
+      cy.contains(
+        `Are you sure you want to permanently remove mainNumber: 00000111111 from Mark Gardner?`
```
Asura5, line 1793: Any reason we can't remove 'permanently' ? It slightly implies we won't be able to re-add them in the future

LBHCallumM: I agree with your point. I can remove it

adamtry (approved):

cypress/integration/work-order/create/create.spec.js
```diff
@@ -1759,6 +1766,301 @@ describe('Raise repair form', () => {
       )
     })
   })
+
+  context('When a user removes a contact detail', () => {
```
Asura5, line 1770: Can we break this in to multiple spec files, it's getting ridiculous....

src/pages/api/contact-details/index.js
```diff
@@ -0,0 +1,46 @@
+import * as HttpStatus from 'http-status-codes'
+import {
+  authoriseServiceAPIRequest,
+  externalAPIRequest,
+} from '@/utils/serviceApiClient'
+
+export default authoriseServiceAPIRequest(async (req, res) => {
+  if (req.method !== 'DELETE' && req.method !== 'PATCH') return
```
Asura5, line 8: This can be refactored into a switch which would allow better extensibility later. This line can be a default case

src/utils/serviceApiClient.js
```diff
@@ -21,6 +21,92 @@ const {
...
+        'Cache-Control',
+        `public,max-age=${CACHE_MAX_AGE_IN_SECONDS}`
+      )
+      response.setHeader('X-Cache', 'HIT')
+
+      return data
+    }
+
+    const cookies = cookie.parse(request.headers.cookie ?? '')
+    const token = cookies[GSSO_TOKEN_NAME]
```
Asura5, line 41: Can we check this token isn't null or whitespace before using it?

src/utils/serviceApiClient.js
```diff
@@ -21,6 +21,92 @@ const {
...
+      Authorization: token,
+      'Content-Type': 'application/json',
+    }
+
+    let { path, ...queryParams } = request.query
+
+    const api = axios.create()
+
+    // Log request
+    api.interceptors.request.use((request) => {
```
Asura5, line 55: Might be worth refactoring this out in to a logging utility method

src/pages/api/contact-details/index.test.js
```diff
@@ -0,0 +1,105 @@
...
+    'x-api-key': REPAIRS_SERVICE_API_KEY,
+    'x-hackney-user': signedCookie,
+    Authorization: signedCookie,
+  }
+
+  test('DELETE request forwards request to correct endpoint', async () => {
+    const contactId = 'aaa'
+    const personId = 'bbb'
+
+    const req = createRequest({
```
Asura5, line 51: I'd change this to createDeleteRequest, otherwise it's ambiguous

src/components/Property/Contacts/TenantsContactTable/TenantPhoneNumber.js
```diff
@@ -0,0 +1,65 @@
+import { useState } from 'react'
+import RemoveContactModal from './modals/RemoveContactModal'
+import SetAsMainModal from './modals/SetAsMainModal'
+import classNames from 'classnames'
+
+const TenantPhoneNumber = ({ phoneNumber, tenant, reloadContacts }) => {
```
Asura5, line 6: As we're now effectively coupling MMH and RepairsHub, how about an ErrorBoundary around these likely-to-break components?

cypress/integration/work-order/create/create.spec.js
LBHCallumM: Sure. Despite my initial reluctance, I'm starting to like using Cucumber for Cypress. These tests are pretty difficult to manage

src/utils/serviceApiClient.js
LBHCallumM: Happy to do refactoring on this file, but I think it would be easier to do in a separate PR. This PR is already quite large

src/pages/api/contact-details/index.test.js
LBHCallumM: Not sure how helpful this is here, because in both scenarios, the same endpoint is being called. However, I have extracted this logic into two separate methods: `createDeleteRequest` and `createPatchRequest`. This does make the code more readable, so I will keep it

src/components/Property/Contacts/TenantsContactTable/TenantPhoneNumber.js
LBHCallumM: What part of this code do you see as likely to break? To summarise the changes:

- Its using the existing integration with the contact details endpoint to display them
- A link to MMH, which is populated by an environment variable
- PATCH and DELETE requests called from within confirmation modals. These contain error handling

I'm not sure what part you think needs an error boundary.

Asura5 (approved):
All comments look good, ignore the ErrorBoundary comment.

PR #795, opened 2023-12-11
Add Operative MobileView to backoffice

LBHCallumM:
## Summary of Changes

Add Operative MobileView to backoffice so we can see what jobs an operative can see.

- Separate currentUser logic from existing MobileWorkingWorkOrdersView component so we can manually pass in an operative in the backoffice
- Add mobile view page to backoffice


adamtry (approved):
Looks good

PR #825, opened 2024-02-12
Re-enable e2e tests

LBHCallumM:
## Summary of Changes

Fix flaky tests + add required tests changes for DRS session logic changes.

- Hotjar was breaking the tests. So only init when not in Cypress env
- Remove any test logic related to setting DRS Session Cookies. In some cases, replace with an intercept to fetch a new session.

### DIFF
--- a/src/components/Meta/index.js
+++ b/src/components/Meta/index.js
@@ -5,7 +5,7 @@ import { useEffect } from 'react'
 const Meta = ({ title }) => {
   const applicationTitle = 'Hackney Repairs Hub'
   useEffect(() => {
-    hotjar.initialize(3085570, 6)
+    // hotjar.initialize(3085570, 6)
   }, [])
 
   return (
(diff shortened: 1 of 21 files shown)


src/components/Meta/index.js
```diff
@@ -5,7 +5,10 @@ import { useEffect } from 'react'
 const Meta = ({ title }) => {
   const applicationTitle = 'Hackney Repairs Hub'
   useEffect(() => {
-    hotjar.initialize(3085570, 6)
+    if (!window.Cypress) {
```
adamtry, line 8: Is there an easy way to also exclude dev and staging?

adamtry (approved):

src/components/Meta/index.js
LBHCallumM: Perhaps. But is there a need to do so?

PR #831, opened 2024-02-23
Add logging to mobile view

LBHCallumM:
## Summary of Changes

Logs what operatives can see on the mobile view.


src/components/WorkOrders/MobileWorkingWorkOrdersView/MobileWorkingWorkOrdersView.js
```diff
@@ -54,28 +84,29 @@ const MobileWorkingWorkOrdersView = ({ currentUser }) => {
...
-              return ''
-            }
-          })()}
-          currentUser={currentUser}
-        />
-      ))
-    )
+    if (workOrders.length === 0) {
+      return <></>
+    }
```
LBHCallumM, line 89: Previous logic displayed '0' when empty

src/components/WorkOrders/MobileWorkingWorkOrdersView/MobileWorkingWorkOrdersView.js
```diff
@@ -9,11 +9,14 @@ import WarningInfoBox from '../../Template/WarningInfoBox'
 import Meta from '../../Meta'
 import { WorkOrder } from '../../../models/workOrder'
 
-const MobileWorkingWorkOrdersView = ({ currentUser }) => {
+const MobileWorkingWorkOrdersView = ({
+  currentUser,
+  loggingEnabled = true,
```
adamtry, line 14: Does this need a feature toggle?

LBHCallumM: its not a feature toggle

src/components/WorkOrders/MobileWorkingWorkOrdersView/MobileWorkingWorkOrdersView.js
```diff
@@ -99,6 +132,32 @@ const MobileWorkingWorkOrdersView = ({ currentUser }) => {
       .slice(0, 1)
   }
 
+  const logWorkOrders = async (
```
adamtry, line 135: This is good info to have and definitely very useful as-is. Could consider adding some timestamps so we won't need to look them up after?

So in progress work orders, started work orders, and current work orders as objects including id, start /end times, and scheduled start / end times.

But if this needs API changes then don't worry about it for now I guess - this is very good as-is

src/components/WorkOrders/MobileWorkingWorkOrdersView/MobileWorkingWorkOrdersView.js
```diff
@@ -99,6 +132,32 @@ const MobileWorkingWorkOrdersView = ({ currentUser }) => {
...
+    }
+
+    try {
+      await frontEndApiRequest({
+        method: 'post',
+        path: '/api/frontend-logging/operative-mobile-view-work-orders',
+        requestData: body,
+      })
+    } catch (error) {
+      // fail silently - user doesnt need to know if logging fails
```
adamtry, line 157: Could consider a console.error here just so we can investigate in the unlikely event of a failure

adamtry (approved):

src/components/WorkOrders/MobileWorkingWorkOrdersView/MobileWorkingWorkOrdersView.js
LBHCallumM: timestamps are set within the api

LBHCallumM: We could just log when returning data from the api. But I think whats more useful is knowing what the frontend sees. So we can see the result of any filtering logic on the frontend

LBHCallumM: If we find this useful, there is no reason we couldnt change the implementation

PR #842, opened 2024-03-27
Staging update

LBHCallumM:


Duslerke (approved):
Looking at the other `contractorName`s above the new one - they follow different casing.
Is the name of the new one actually meant to be all caps?

LBHCallumM:
> Looking at the other `contractorName`s above the new one - they follow different casing. Is the name of the new one actually meant to be all caps?

Probably not. But doesnt matter other than αєѕтнєтι¢ѕ. Can always change it later

PR #848, opened 2024-04-04
Remove unused dependencies

LBHCallumM:
### Description of change

Removed unused dependencies based on the output of depcheck

```
> npx depcheck
npx: installed 128 in 9.431s
Unused dependencies
* dotenv
* path
* universal-cookie
Unused devDependencies
* @babel/plugin-proposal-class-properties
* @babel/plugin-transform-private-methods
* cypress-dotenv
Missing dependencies
* src: .\__test__\search.test.js
* factories: .\src\utils\headerLinks.test.js
* prop-types: .\src\pages\login.js
* dedent: .\cypress\integration\work-order\create\create-with-adding-multiple-sors.spec.js
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,6 @@
     "sass": "^1.29.0",
     "serve-static": "^1.14.1",
     "serverless-http": "^2.6.0",
-    "universal-cookie": "^4.0.4",
     "uuid": "^8.3.2",
     "xmldom": "^0.6.0",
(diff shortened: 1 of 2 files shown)


Duslerke (commented):
Interesting that the e2e tests fail.

Could it be that some of these removed dependencies aren't used directly by the project, but instead are used by another dependency indirectly?

If so, maybe `npm install` would resolve that?

PR #849, opened 2024-04-10
Remove Purdy from multi trade contractors list

LBHCallumM:


Duslerke (approved):

src/utils/constants.js
```diff
@@ -1,13 +1,13 @@
 export const MULTITRADE_ENABLED_CONTRACTORS = [
-  'PCL',
+  'FPM',
+  // 'PCL',
```
humulla, line 3: Remove it, if we don't need it

PR #872, opened 2024-07-25
Add tests for photos tab

LBHCallumM:
## Summary of Changes

Add E2E tests for photos tab. Plus add snapshot tests for each new component


src/components/WorkOrder/Photos/PhotoGroupDescriptionForm.js
```diff
@@ -0,0 +1,106 @@
...
+      />
+    )
+  }
+
+  return (
+    <UpdateDescriptionButton description={description} showForm={showForm} />
+  )
+}
+
+export const UpdateDescriptionButton = ({ showForm, description }) => {
```
adamtry, line 40: Does this need to be public + tested?
I'd think it could be private and remain just part of the PhotoGroupDescriptionForm component, tested as part of that component also

adamtry (approved):

src/components/WorkOrder/Photos/PhotoGroupDescriptionForm.js
LBHCallumM: I was actually testing the parent. but i found it was easier to test the children. should probably refactor it

PR #885, opened 2024-08-01
Remove image, and point file upload to google drive

LBHCallumM:
### Description of change

- Removed image
- File can be hosted on Google Drive


Duslerke (commented):
I could have been merely a bundling or configuration issue.
I'm pretty sure I recall that hosting images from /public was supported last time I worked with next. :thinking:

PR #894, opened 2024-08-15
Fix failing search tests

LBHCallumM:


LBHCallumM:
> Good old race conditions :D

This is why these tests are so hard to fix XD

PR #919, opened 2024-09-27
Fix date issue on work order notes

LBHCallumM:
### Description of change

The closedDate property from getWorkOrder endpoint is actually spelled `closedDated`. 

This caused the date to be missing from the note.

### DIFF
--- a/src/components/WorkOrder/Notes/NoteContent/CompletedNoteContent/index.test.js
+++ b/src/components/WorkOrder/Notes/NoteContent/CompletedNoteContent/index.test.js
@@ -4,7 +4,7 @@ import CompletedNoteContent from '.'
 describe('CompletedNoteContent component', () => {
   it('should match snapshot when user comment is empty', () => {
     const workOrder = {
-      closedDate: new Date('2024-09-26T12:34:56'), // Replace with actual date fixture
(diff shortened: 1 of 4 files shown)


Duslerke (commented):
Why not fix the spelling mistake in the source instead?

LBHCallumM:
> Why not fix the spelling mistake in the source? _(instead of propagating it into a wider codebase)_
> 
> Seems like it would be a more wholesome fix.

where

LBHCallumM:
> Why not fix the spelling mistake in the source? _(instead of propagating it into a wider codebase)_
> 
> Seems like it would be a more wholesome fix.

Making a small change, vs two prs and extensive changes accross the frontend? hmmm

Duslerke:
> > Why not fix the spelling mistake in the source? _(instead of propagating it into a wider codebase)_
> > Seems like it would be a more wholesome fix.
> 
> Making a small change, vs two prs and extensive changes accross the frontend? hmmm

For where: wherever the data is coming from (if reasonable)
2 PRs: It will make the overall code cleaner for everyone, plus this will remove the ambiguity as to what this key actually means now. It's unclear whether it's meant to mean: "closed date" or "close dated". :grin:

adamtry (approved):

PR #925, opened 2024-10-07
Add feature toggle for follow-on functionality

LBHCallumM:
## Summary of Changes

Add feature toggle to disable follow-on functionality.

I wanted to clone the E2E tests so I can test when the FT was enabled and disabled. Using an enviornment variable made this impossible due to how NextJs works.

The solution was to add a featureToggle endpoint. The endpoint simply reads the environment variable. This made testing much easier as I can intercept the request to the feature toggle endpoint. 

I decided against using the configuration API because it was too confusing. Whereas setting up a new API endpoint in nextjs is increadibly simple.

### Additional changes

- I had to update a few components to typescript, as they were causing eslint to fail

note to self
```yml
      FOLLOW_ON_FUNCTIONALITY_ENABLED: ${ssm:/repairs-hub/${self:provider.stage}/repairs-hub-enable-followon-functionality}

```

### DIFF
--- a/cypress/e2e/work-order/attend/close-by-proxy-with-followons-enabled.cy.js
+++ b/cypress/e2e/work-order/attend/close-by-proxy-with-followons-enabled.cy.js
@@ -1,9 +1,9 @@
 /// <reference types="cypress" />
 import 'cypress-audit/commands'
 
-describe('Closing a work order on behalf of an operative', () => {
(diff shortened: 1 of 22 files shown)


adamtry (approved):
I'll take a feature toggle endpoint on the API over the configuration API any day

PR #928, opened 2024-10-11
Add retry mechanism to s3 upload

LBHCallumM:
## Summary of Changes

Add retry mechanism to S3 upload. Specifically, uploading the files to S3 using presigned urls.

For each file, an upload will be retried 3 times with exponential backoff. 

https://yarnpkg.com/package?q=retry&name=retry

## Additional changes

- Fixed the error handling when any file upload fails. CompleteUpload should only be called on success. 
- Refactored `completionDateTime` and converted to TS
- Replaced `cy.clock()` with cypress environment variables in` isCurrentTimeOperativeOvertime`. Modifying the system time was breaking the retry mechanism.


Duslerke (dismissed):
Sorry not much time today to do a review you've asked, but I'll just leave these couple thoughts (not expecting any answers) on what I'd be looking for in review:

The 2 things I'd ask:
1. are the retries spaced out enough so it's not one after another in milliseconds? _(what are the defaults? it seems this retry.operation call is falling upon default vals)_ If not, each retry will likely experience the same circumstances that caused the original request to fail.
2. if the retries are in the seconds range, is the user aware that they should wait around for that long (and not close out/refresh the window if they're unhappy with no success/page redirect feedback)

LBHCallumM:
> Sorry not much time today to do a review you've asked, but I'll just leave these couple thoughts (not expecting any answers) on what I'd be looking for in review:
> 
> The 2 things I'd ask:
> 
> 1. are the retries spaced out enough so it's not one after another in milliseconds? _(what are the defaults? it seems this retry.operation call is falling upon default vals)_ If not, each retry will likely experience the same circumstances that caused the original request to fail.
> 2. if the retries are in the seconds range, is the user aware that they should wait around for that long (and not close out/refresh the window if they're unhappy with no success/page redirect feedback)

The library handles the exponential timeout. I think its about a second, and doubles each time.

This may annoy the user, but they are uploading a file. But not sure what alternative there is.

src/components/WorkOrder/Photos/hooks/uploadFiles/faultTolerantRequest.ts
```diff
@@ -0,0 +1,36 @@
+import retry from 'retry'
+
+const DEFAULT_NUMBER_OF_RETRIES = 3
+
+const faultTolerantRequest = async (
```
adamtry, line 5: I'd be confused if I saw this without context in future

```suggestion
/**
 * Returns a promise that tries to successfully resolve
 * a specified amount of times with exponential backoff
 */
const faultTolerantRequest = async (
```

adamtry (approved):

PR #937, opened 2024-10-18
Remove cache from requests

LBHCallumM:
## Summary of Changes

All requests sent from RH are intercepted by Nextjs API and cached. But given that next is running on Lambda, the caching is kinda pointless (at least while stored in memory).

I think that the caching could be causing the 500 errors not to clear for some users. So I am removing it.

It should also reduce the complexity a little bit.


adamtry (approved):
Looks good, just needs testing in dev/staging to verify this isn't affecting performance

PR #944, opened 2024-10-23
Add OOH gas check to can be scheduled

LBHCallumM:
## Summary of changes

Prevent OOH gas jobs from raising an appointment with DRS


Duslerke (approved):
Seems alright, though it's odd to see these PascalCased variable names.

Any reason for going against JavaScript convention?

LBHCallumM:
> Seems alright, though it's odd to see these PascalCased variable names.
> 
> Any reason for not going with the JavaScript convention? (PascalCase is reserved for class names I think)

I copied the code from .net. forgot to change it

PR #948, opened 2024-10-24
convert search file to typescript

AlexpdHackney:
### Description of change

Change the search page to Typescript.


tsconfig.json
```diff
@@ -25,6 +25,6 @@
       "@/utils/*": ["src/utils/*"]
     }
   },
-  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx"],
+  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", "src/components/Search/Search.test.tsx", "src/components/Search/Search.tsx"],
```
Duslerke, line 28: These 2 patterns seem like they are already part of the `"**/*.tsx"` pattern.

AlexpdHackney: @Duslerke This should be fixed now!

src/components/Search/Search.tsx
```diff
@@ -10,10 +10,22 @@ import Meta from '../Meta'
...
-const Search = ({ query }) => {
+interface SearchProps {
+  query?: {
+    searchText?: string
+    pageNumber?: string
+  }
+}
+
+interface PropertyData {
+  total: string
```
Duslerke, line 21: Based on the line 90, I'm guessing it actually comes through as string from the API?
```js
setSearchHitTotal(parseInt(propertiesData.total))
```

src/components/Search/Search.tsx
```diff
@@ -10,10 +10,22 @@ import Meta from '../Meta'
 import { canSearchForProperty } from '@/utils/userPermissions'
 import { PropertyListItem } from '@/models/propertyListItem'
 
-const Search = ({ query }) => {
+interface SearchProps {
+  query?: {
+    searchText?: string
+    pageNumber?: string
```
Duslerke, line 16: Guess it's the same issue with the pageNumber:
```ts
...(pageNumber && { pageNumber: parseInt(pageNumber) }),
```

Duslerke (approved):
Looking at the previous code, all is probably fine.

However, I raised the questions about the data types just in case.

src/components/Search/Search.tsx
Duslerke: Hmm the API seems to be returning a `long` [[link to model](https://github.com/LBHackney-IT/repairs-api-dotnet/blob/423055144f1f7df6a8942c94f3d2c0906c0bca9b/RepairsApi/V2/Boundary/Response/SearchForPropertiesResponse.cs#L10)].

src/components/Search/Search.tsx
Duslerke: The repairs API seems to treat this as an integer [[link to model](https://github.com/LBHackney-IT/repairs-api-dotnet/blob/423055144f1f7df6a8942c94f3d2c0906c0bca9b/RepairsApi/V2/Boundary/Request/SearchForPropertiesRequest.cs#L18)].

src/components/Search/Search.tsx
LBHCallumM: its a number

LBHCallumM (approved):

PR #953, opened 2024-11-05
Staging update

LBHCallumM:


src/components/Search/Search.tsx
```diff
@@ -10,10 +10,22 @@ import Meta from '../Meta'
...
-const Search = ({ query }) => {
+interface SearchProps {
+  query?: {
+    searchText?: string
+    pageNumber?: string
+  }
+}
+
+interface PropertyData {
+  total: string
```
Duslerke, line 21: So long as `total` as string instead of a number doesn't cause any issues , then it's fine.

src/components/Search/Search.tsx
```diff
@@ -10,10 +10,22 @@ import Meta from '../Meta'
 import { canSearchForProperty } from '@/utils/userPermissions'
 import { PropertyListItem } from '@/models/propertyListItem'
 
-const Search = ({ query }) => {
+interface SearchProps {
+  query?: {
+    searchText?: string
+    pageNumber?: string
```
Duslerke, line 16: Seems like it was decided to go with the `pageNumber` as a string rather than a number. So long as everyone is happy about it, I can approve.

Duslerke (approved):

PR #960, opened 2024-11-14
Parallelise Cypress tests in CI

spikeheap:
### Description of change

This change introduces [CircleCI parallelism](https://circleci.com/docs/parallelism-faster-jobs/), to split the Cypress end to end tests and run them on multiple workers at the same time.

This leverages CircleCI's test splitter, which records the test duration and balances the split so each worker should complete at roughly the same time.

Before this change the end to end step took about 9.5 minutes. 

After the change it's down to just below 5 minutes. CircleCI guesses this is a 60% build time improvement on average 🤩 .

Screenshot 2024-11-14 at 15 11 (Brave Browser)@2x

In order for CircleCI to understand how long a test is running, we need to [output test results in a standard JUnit format](https://docs.cypress.io/app/tooling/reporters). This PR also includes that change, which logs to `cypress/results` automatically and instructs Circle to [slurp results from that directory](https://circleci.com/docs/collect-test-data/#cypress). This also means we'll gain insights into test execution times and test flakiness within CircleCI.

Screenshot 2024-11-14 at 16 46 (Brave Browser)@2x

The test splitting code is loosely based on [this example](https://circleci.com/docs/troubleshoot-test-splitting/), adapted for Cypress.

### Automated test checklist

- [x] Unit tests (Jest)

### Decisions

1. Parallelism was set to 4. Setting it to 2 yielded a significant improvement (around 6.5 minutes), and 4 shaved another 90 seconds off. I didn't take it further because the setup time is now a significant component so we'll see diminishing returns.


LBHCallumM (approved):
Amazing work!

Duslerke (approved):

PR #984, opened 2024-12-04
latest changes updated

AlexpdHackney:
### Description of change

Update the latest changes to reflect new changes.


src/components/LatestChanges/index.js
```diff
@@ -1,171 +1,135 @@
...
-            Library.
-          </li>
-        </ul>
+
+        <Image
+          src={closeWorkOrder}
+          width={300}
+          height={300}
+          alt="Close work order"
+        />
```
LBHCallumM, line 45: I would recomend setting both width and height. I would set one or the other, so the image is responsive

src/components/LatestChanges/index.js
```diff
@@ -1,171 +1,135 @@
+import Image from 'next/image'
 import BackButton from '../Layout/BackButton'
+import closeWorkOrder from '../../../public/images/closeworkorder.jpg'
+import detailsOfFurtherWork from '../../../public/images/detailsoffurtherwork.jpg'
+import additionalNotes from '../../../public/images/additionalnotes.jpg'
```
Duslerke, line 5: Does this grab the file path, or the binary itself?

AlexpdHackney: As far as I'm aware it grabs the file path

src/components/LatestChanges/index.js
AlexpdHackney: done

LBHCallumM (approved):
lgtm

PR #995, opened 2024-12-12
Remove explicit serverless version from config

spikeheap:
Serverless is only invoked through CI. Upgrading to v4 failed because we needed to update in two places (the npm install and serverless.yml frameworkVersion attribute.

Other repositories don't set this explicitly, and as we don't run serverless locally we don't need this to enforce consistency. ;

### DIFF
--- a/serverless.yml
+++ b/serverless.yml
@@ -1,7 +1,5 @@
 service: repairs-hub-frontend
 
-frameworkVersion: '2'
-
 provider:
   name: aws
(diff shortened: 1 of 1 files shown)


spikeheap:
If we'd prefer to make major version upgrades explicit this PR could change `2` to `4`. I opted to make it consistent with other Hackney serverless setups, but I don't have any preference :)

LBHCallumM (approved):

PR #1022, opened 2025-03-06
Improve other follow on trade flow 

AlexpdHackney:
#### Context:

We have issues when the operatives are selecting 'other' and then utilising the free text input field that it automatically goes to reactive repairs when there is likelihood it should go to another inbox - ie gas or external works. 

#### Solution: 

Instead of adding endlessly to the list of options we have decided to make the ‘other’ input field smarter, turning it into an intuitive drop down once the user starts typing - providing options that they can select. 

### Story Link

https://hackney.atlassian.net/jira/software/projects/HPT/boards/116?selectedIssue=HPT-640

### Automated test checklist

### Decisions

I chose to use the function underneath in the FollowOnRequestDifferentTradesForm file for two reasons. Firstly I tried to figure out how I could use the methods and classes in utils\helpers\filter to achieve the same filter but I couldn't figure it out. Secondly I thought that this function won't be used much anywhere else so I decided to keep it in the file rather than move it to a utils folder. This can all be changed if needed. 

image

#### Page to visit and options to select:

image

image
 
#### How it initially looks:

image

#### Search for trades with 'A' in them': 

image

#### Select 'Asbestos':

image

#### Enter text not in dropdown:

image


src/types/requests/types.ts
```diff
@@ -1,4 +1,4 @@
-export type ApiResponseType<T> = {
+export type ApiResponseType<T = unknown> = {
```
LBHCallumM, line 1: whats unknown for?

src/components/WorkOrders/CloseWorkOrderFormComponents/FollowOnRequestDifferentTradesForm.js
```diff
@@ -1,11 +1,61 @@
...
+import { getTrades } from '@/root/src/utils/requests/trades'
 
 const FollowOnRequestDifferentTradesForm = (props) => {
-  const { register, requiredFollowOnTrades } = props
+  const { register, requiredFollowOnTrades, watch, errors } = props
+
+  const [trades, setTrades] = useState([])
+  const [filteredTrades, setFilteredTrades] = useState([])
+  const [error, setError] = useState(null)
+  const maxLength = 100
```
LBHCallumM, line 14: Can this constant me moved outside of the render function

src/components/WorkOrders/CloseWorkOrderFormComponents/FollowOnRequestDifferentTradesForm.js
```diff
@@ -1,11 +1,61 @@
...
+      'PN',
+      'PL',
+      'RF',
+      'SC',
+      'UP',
+      'SV',
+    ]
+    const filteredTrades = trades.filter(
+      (trade) => !codesToFilter.includes(trade.key)
+    )
```
LBHCallumM, line 55: See if you can convert this to a `Set<string>`. It would be a lol cleaner. Secondly, this codes are a constant, so I would remove them from the render function (above the component code).

src/components/WorkOrders/CloseWorkOrderFormComponents/FollowOnRequestDifferentTradesForm.js
```diff
@@ -21,6 +71,25 @@ const FollowOnRequestDifferentTradesForm = (props) => {
...
+        <>
+          <DataList
+            label="Please specify"
+            name="otherTrade"
+            options={filteredTrades.map((trade) => trade.description)}
+            register={register}
+            hint="Select or type a trade"
+            widthClass="govuk-!-width-full"
+            error={errors.otherTrade}
+            maxLength={100}
```
LBHCallumM, line 84: If were using a maxLength variable, can we also use it here

src/types/requests/types.ts
AlexpdHackney: Changed!

src/components/WorkOrders/CloseWorkOrderFormComponents/FollowOnRequestDifferentTradesForm.js
AlexpdHackney: Done

LBHCallumM (approved):

PR #1029, opened 2025-03-31
Add estimated duration dropdown

AlexpdHackney:
#### Context 

When designing the solution it was initially decided that we didn’t want the operatives placing estimations on the form - this was signed off by the exec. Though since going live we have had feedback it would be helpful - especially from plumbing who provide the highest number of follow ons as it helps inform the SORs and also plan the work. 

##Acceptance Criteria 

Ability to add information on the estimated time needed to complete a task on the follow on form 

After the ‘Describe work required’ input box but before the materials required 

Mandatory if we have an unknown option if using the radio buttons 

Build both UX versions and push to dev - radio buttons and drop down - so we can test which is the most usable

### Story Link

https://hackney.atlassian.net/jira/software/projects/HPT/boards/116?selectedIssue=HPT-630

### Automated test checklist

### Known issues

According to whichever option is chosen the tests and the components used will need to be altered. I will also need to get rid of the Radio or Dropdown label and buttons. 

##### Mobile: 

image

image

image

##### Desktop: 

image

image

image

### DIFF
--- a/cypress/e2e/work-order/attend/close-by-proxy-with-followons-enabled.cy.js
+++ b/cypress/e2e/work-order/attend/close-by-proxy-with-followons-enabled.cy.js
@@ -2858,7 +2858,7 @@ describe('Closing a work order on behalf of an operative - When follow-ons are e
           'not.exist'
         )
         // add estimated duration of work - error should disappear]
-        cy.get('[data-testid="radiobutton"]').click()
+        cy.get('[data-testid="radioButton"]').click()
         cy.get('input[data-testid="estimatedDuration"]').first().check()
         cy.contains('Select a duration').should('not.exist')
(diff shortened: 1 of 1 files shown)


cypress/e2e/work-order/attend/close-with-followons-enabled.cy.js
```diff
@@ -946,7 +950,7 @@ describe('Closing my own work order - When follow-ons are enabled', () => {
     cy.contains('Work order 10000621 successfully closed')
   })
 
-  it('throws an error when other trade is not submitted properly', () => {
+  it.only('throws an error when other trade is not submitted properly', () => {
```
LBHCallumM, line 953: intentional?

LBHCallumM (approved):

PR #1035, opened 2025-04-07
Make follow on look nicer

AlexpdHackney:
#### Context 

This is due to my misunderstanding of the desktop process to close a job, the different google groups and processes involved. 

When testing the follow on form on staging I was taken by how unattractive and unclear the UX is for the follow on form. Looking back to some original designs I think it would be beneficial to implement the format that was conceptualised: 

#### Acceptance 

Have everything as one list apart from the other trade checkboxes 

#### Initial design: 

image

**Grey background is a must and is only for desktop view**

### Story Link

https://hackney.atlassian.net/jira/software/projects/HPT/boards/116?selectedIssue=HPT-650

### How it looks

image

### With other not selected:

image

### With other selected: 

image

#### With radio estimated duration selected: 

image

#### With dropdown estimated duration selected:

image

### DIFF
--- a/src/components/WorkOrders/CloseWorkOrderFormComponents/FollowOnRequestTypeOfWorkForm.js
+++ b/src/components/WorkOrders/CloseWorkOrderFormComponents/FollowOnRequestTypeOfWorkForm.js
@@ -15,6 +15,7 @@ const FollowOnRequestTypeOfWorkForm = (props) => {
     watch,
     followOnData,
     hasWhiteBackground,
+    isGrid,
   } = props
 
   const selectedFurtherWorkRequired =
@@ -136,6 +137,7 @@ const FollowOnRequestTypeOfWorkForm = (props) => {
                   requiredFollowOnTrades={
                     followOnData?.requiredFollowOnTrades ?? []
                   }
+                  isGrid={isGrid}
                 />
               </>
             }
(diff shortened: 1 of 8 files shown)


Duslerke (commented):
Looks nice. The break of tickboxes into 2 columns looks much better.

Though, regarding the rounding of corners. It does not look like GovUK design pattern Hackney ought to be following. :thinking: Or is the 1st photo not "how it looks after the changes"? The "How it looks" is it "before" or "after"?

@LBHCallumM, not 100% sure _(been a while out of FE)_ if this is a requirement though, but I think we should be following the GovUK styles, right?

src/components/WorkOrders/CloseWorkOrderFormComponents/FollowOnRequestTypeOfWorkForm.js
```diff
@@ -14,6 +14,8 @@ const FollowOnRequestTypeOfWorkForm = (props) => {
     clearErrors,
     watch,
     followOnData,
+    hasWhiteBackground,
+    isGrid,
```
LBHCallumM, line 18: Adding these style related props is going to become unmaintainable quickly. Could we instead use a class prop, which combines them all.

For example:
```
import cx from 'classnames'
...
<
  class={cx("checkbox-classes", classProp)}
/>



```

LBHCallumM (approved):

PR #1048, opened 2025-04-22
Add follow-on section to CreateWorkOrder form

LBHCallumM:
## Summary of Changes

Add a follow-on section to raise the work order form. This section is only available to users in the `FollowOn-Admin` group.

image

It fetches the 5 most recent jobs for that property using the preexisting work order history endpoint.

image

The `look up work order number` link opens the search page in a new tab. This was a deliberate design choice, as users are used to opening multiple windows.


src/components/Property/RaiseWorkOrder/TradeContractorRateScheduleItemView.js
```diff
@@ -131,7 +131,6 @@ const TradeContractorRateScheduleItemView = ({
       } else {
         await prepareSORData(contractorRef, tradeCode)
       }
-      console.log(`Contractor: ${contractorRef}`)
```
LBHCallumM, line 134: this log has bugged me for ages

src/components/Property/RaiseWorkOrder/RaiseWorkOrderForm.js
```diff
@@ -210,6 +228,15 @@ const RaiseWorkOrderForm = ({
...
             onSubmit={handleSubmit(onSubmit)}
           >
+            {canAssignFollowOnRelationship(user) && (
+              <RaiseWorkOrderFollowOn
+                register={register}
+                errors={errors}
+                propertyReference={propertyReference}
+                watch={watch}
+              />
+            )}
```
LBHCallumM, line 238: Conditionally show on the form

src/components/Property/RaiseWorkOrder/RaiseWorkOrderFollowOn/FollowOnLookup.tsx
```diff
@@ -0,0 +1,96 @@
...
+      console.error('An error has occured:', e.response)
+      setError(
+        `Oops an error occurred with error status: ${
+          e.response?.status
+        } with message: ${JSON.stringify(e.response?.data?.message)}`
+      )
+    }
+
+    setIsLoading(false)
+  }
```
LBHCallumM, line 56: Fetch 5 most recent work orders form the workOrderHistory endpoint

src/components/Property/RaiseWorkOrder/RaiseWorkOrderFollowOn/FollowOnLookup.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+  const [workOrders, setWorkOrders] = useState<WorkOrder[]>([])
+  const [isLoading, setIsLoading] = useState<boolean>(false)
+  const [error, setError] = useState<string | null>(null)
+
+  const fetchRelatedWorkOrders = async () => {
+    if (workOrders.length > 0) {
+      // we dont need to fetch new data
+      return
+    }
```
Duslerke, line 26: Is it possible that we already have some workorders, but not all?

src/components/Property/RaiseWorkOrder/RaiseWorkOrderFollowOn/FollowOnLookup.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+    fetchRelatedWorkOrders()
+  }, [visible])
+
+  if (!visible) return null
+
+  if (isLoading) {
+    return <SpinnerWithLabel label="Fetching recent work orders" />
+  }
+
```
Duslerke, line 75: If you were to move up lines 70 and 72 above the `useEffect` would you still need lines 64 and 65?

Duslerke (approved):

src/components/Property/RaiseWorkOrder/RaiseWorkOrderFollowOn/FollowOnLookup.tsx
LBHCallumM: Not entirely sure what you mean. This only fetches the 5 most recent work orders as an auto suggestion. The user will be pasting in a work order reference anyway

src/components/Property/RaiseWorkOrder/RaiseWorkOrderFollowOn/FollowOnLookup.tsx
LBHCallumM: Hooks need to be at the top of components. You cant call them conditionally. 

https://dev.to/collegewap/fix-react-hook-useeffect-is-called-conditionally-25m5

src/components/Property/RaiseWorkOrder/RaiseWorkOrderFollowOn/FollowOnLookup.tsx
Duslerke: I was thinking whether it's possible to have that we have 3WOs loaded, when 5 are available or smth.
If yes, this condition wouldn't load the remaining 2.

But from what you're saying that's not possible, so all good.

src/components/Property/RaiseWorkOrder/RaiseWorkOrderFollowOn/FollowOnLookup.tsx
Duslerke: I see, thanks for clearing that up!

src/components/Property/RaiseWorkOrder/RaiseWorkOrderFollowOn/FollowOnLookup.tsx
LBHCallumM: The number of related work orders arent going to change that frequently. Its not a race condition, if thats what youre suggesting

PR #1064, opened 2025-05-09
Update work order components to typescript

LBHCallumM:
Update work order model and related components to typescript.

This is to make upcoming changes easier


src/components/WorkOrder/__snapshots__/MobileWorkingWorkOrder.test.js.snap
```diff
@@ -2325,9 +2309,7 @@ exports[`MobileWorkingWorkOrder component with startTime when the workOrder has
     <h2
       class="lbh-heading-h2"
     >
-      Friday 3 September
-      <br />
-      08:00 – 13:00
+      No appointment
```
adamtry, line 2312: Don't these snapshot changes indicate a bug?

LBHCallumM: they would indeed

adamtry (approved):

PR #1066, opened 2025-05-09
Extract getWorkOrder logic into a single method

LBHCallumM:
I'm currently trying to separate fetching appointment details from fetching work order details. This will enable more resilience when DRS goes down.

However, the PR became very large. The risk of it breaking something became too high.

This PR extracts all the getWorkOrder logic into a single method. In future iterations, it will allow more control and easier feature toggling when updating this feature.

## Key changes

The GetWorkOrder method returns a response object, rather than throwing errors directly. 

image

To integrate well with preexisting error handling, I've introduced the error class `APIResponseError`. This custom error behaves identically to the base Error class and allows checking for that type of error.

image

As more requests are refactored with this approach, this type check could be removed.

### DIFF
--- a/cypress/e2e/search/search.cy.js
+++ b/cypress/e2e/search/search.cy.js
@@ -152,7 +152,7 @@ describe('Search', () => {
           cy.wait(['@repairs_with_error', '@workOrderTasksRequest'])
         })
 
-        it('Displays an error message', () => {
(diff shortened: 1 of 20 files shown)


adamtry (approved):
All looks good

Duslerke (approved):

PR #1079, opened 2025-05-20
Update Mobile Variation form

LBHCallumM:
## Summary of changes

Recent changes to vary limits have highlighted how terrible the UX is when operatives vary a work order.

If they vary a job above their limit, the status is set to "Pending Variation Approval". They are not told about this, and whenever they try to complete or edit the work order, they see this generic error.

image

This PR adds various validation, and uses the same variation form used on desktop. The UX is not perfect, it's intended as a quick fix the the horrendous ux.

image

PR #1088, opened 2025-05-21
Revert "Update Mobile Variation form"

LBHCallumM:
Reverts LBHackney-IT/repairs-hub-frontend#1079


Duslerke (approved):
Why revert? Did it break prod?

PR #1126, opened 2025-06-20
Remember data when photo upload fails

adamtry:
[HPT-763](https://hackney.atlassian.net/browse/HPT-763) Allow the app to remember the data even if the photo upload fails

Uses the `react-hook-form-persist` package to back-up the mobile work order form to browser local storage, so that if there's an error with the file upload, the form can re-set to the previous values set by the user.

This change is to mitigate time lost by operatives after photo upload fails (e.g. due to internet disconnection) and the resulting frustration, which should also reduce the risk of mandatory photos.

As a bonus, the data is also preserved if the operative exits the app and returns, refreshes the page, etc, however in this case the photos are not preserved because they aren't supported in `react-hook-form-persist` and are preserved using React state instead.

image

[HPT-763]: https://hackney.atlassian.net/browse/HPT-763?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/WorkOrders/CloseWorkOrderFormComponents/CloseWorkOrderFormReasonForClosing.tsx
```diff
@@ -38,25 +37,33 @@ const CloseWorkOrderFormReasonForClosing = (props: Props) => {
...
-        ) : null,
-      }))}
+          ),
+        },
+        {
+          text: noAccess.text,
+          value: noAccess.value,
+          defaultChecked: noAccess.value === reason,
+        },
+      ]}
```
adamtry, line 66: Had quite a few bugs with this with the react-hook-form that were a fair bit easier to troubleshoot without the `.map`

src/components/WorkOrder/MobileWorkingWorkOrderView.js
```diff
@@ -261,11 +271,21 @@ const MobileWorkingWorkOrderView = ({ workOrderReference }) => {
           isNoAccess ? 'closed with no access' : 'closed'
         }`
       )
+      for (let i = 0; i < localStorage.length; i++) {
+        const key = localStorage.key(i)
+        if (key && key.startsWith('closeWorkOrder')) {
+          localStorage.removeItem(key)
+        }
+      } // Clear cached form data
```
LBHCallumM, line 279: could we move this to a clear cache method

adamtry: I gave that a go but it feels a bit premature - that could help more if we need to do something similar with the desktop work order close form also

PR #1128, opened 2025-06-23
Fetch data with react query

AlexpdHackney:
### Description of change

I have decided to use useQuery to fetch contract data as it can cache it, thus making rendering faster on revisit. It will also allow smooth filtering later on.

### DIFF
--- a/src/pages/_app.js
+++ b/src/pages/_app.js
@@ -4,6 +4,7 @@ import App from 'next/app'
 import Layout from '@/components/Layout'
 import AccessDenied from '@/components/AccessDenied'
 import * as Sentry from '@sentry/nextjs'
+import { QueryClient, QueryClientProvider } from 'react-query'
 
 import {
   isAuthorised,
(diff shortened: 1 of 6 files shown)


src/pages/_app.js
```diff
@@ -38,16 +41,18 @@ class MyApp extends App {
...
+            <Layout serviceName="Repairs Hub">
+              <Meta />
+
+              <ComponentToRender
+                {...pageProps}
+                userDetails={this.props.userDetails}
+              />
+            </Layout>
+          </UserContext.Provider>
+        </QueryClientProvider>
```
LBHCallumM, line 55: Like I said above, this should be limited to backoffice.

LBHCallumM: All of the frontend requests use Axios. It would add more complexity by varying how we fetch this data. Im not opposed to trying something different, but we should think about the implications before we roll it out everywhere.

PR #1133, opened 2025-07-01
Sor-code-search

AlexpdHackney:
### Description of change

Add sor code search to the front end api request. Also created a search bar to allow the user to search for contracts related to a contractor with a specific sor code.

### Story Link

https://link-to-issue

### Automated test checklist

- [ ] Unit tests (Jest)
- [ ] Integration tests (Cypress)

### Decisions [OPTIONAL]

If you had to make any decisions between different ways of doing things, what where they and why?

### Known issues [OPTIONAL]

Things you know need further follow on work but aren't in scope of this issue

1. issue1
2. issue2

### DIFF
--- a/src/components/BackOffice/ContractsDashboard/Contractor/ContractorView.tsx
+++ b/src/components/BackOffice/ContractsDashboard/Contractor/ContractorView.tsx
@@ -1,8 +1,10 @@
 import { useQuery } from 'react-query'
+import { useMemo, useState } from 'react'
 import { fetchContracts } from '../../requests'
 
 import Layout from '../../Layout'
 import Spinner from '../../../Spinner'
+import SorSearch from '../SorSearch'
 
 import ContractListItems from '../Contract/ContractListItems'
 
@@ -15,6 +17,8 @@ const ContractorView = ({
   contractorReference,
   contractorName,
 }: ContractorViewProps) => {
+  const [sorCode, setSorCode] = useState<string>(null)
+
   const {
     data: activeContracts,
     isLoading: activeContractsLoading,
@@ -38,8 +42,28 @@ const ContractorView = ({
     ],
     () => fetchContracts(false, contractorReference)
   )
+  const {
+    data: contractsWithSorCode,
+    isLoading: sorContractsIsLoading,
+    refetch: refetchSorContracts,
+    error: sorContractsError,
+  } = useQuery(
+    ['sorContracts', contractorReference, sorCode?.toLocaleUpperCase()],
+    () =>
+      fetchContracts(null, contractorReference, sorCode?.toLocaleUpperCase()),
+    {
+      enabled: false,
+    }
+  )
+
+  const descendingDateContractsWithSorCode = useMemo(
+    () => (contractsWithSorCode ? [...contractsWithSorCode].reverse() : null),
+    [contractsWithSorCode]
+  )
+
   const activeContractsError = activeError as Error | null
   const inactiveContractsError = inactiveError as Error | null
+  const contractsWithSorCodeError = sorContractsError as Error | null
 
   const relativeInactiveContracts = inactiveContracts?.filter(
     (contract) =>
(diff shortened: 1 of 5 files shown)


src/components/BackOffice/ContractsDashboard/SorSearch.tsx
AlexpdHackney, line 1: I chose to create a new component rather than use the Search from forms as there were different requirements that I wanted for it Furthermore, this is for the back office which is a separate part of the app.

src/components/BackOffice/ContractsDashboard/SorSearch.tsx
```diff
@@ -0,0 +1,100 @@
...
+        ) : (
+          <>
+            {contracts?.length > 0 && (
+              <ContractListItems
+                contracts={contracts}
+                heading={`${sorCode} found in the following ${contractorName} contracts`}
+                page="sorSearch"
+              />
+            )}
+            {contracts?.length === 0 && (
```
Duslerke, line 76: the fact that this has a `?`, implies it can become `null`.

Duslerke: Meaning this condition won't trigger when contracts are null

src/components/BackOffice/ContractsDashboard/SorSearch.tsx
```diff
@@ -0,0 +1,100 @@
...
+              </p>
+            )}
+            {error && (
+              <ErrorMessage
+                label={
+                  error instanceof Error
+                    ? error.message
+                    : typeof error === 'string'
+                    ? error
+                    : 'An unexpected error occurred'
```
Duslerke, line 89: This is seems so generic that it could be a helper function.

Duslerke (dismissed):

src/components/BackOffice/ContractsDashboard/Contractor/ContractorView.tsx
```diff
@@ -38,15 +42,39 @@ const ContractorView = ({
...
+    ['sorContracts', contractorReference, sorCode?.toLocaleUpperCase()],
+    () =>
+      fetchContracts(null, contractorReference, sorCode?.toLocaleUpperCase()),
+    {
+      enabled: false,
+    }
+  )
+
+  const descendingDateContractsWithSorCode = useMemo(
+    () => (contractsWithSorCode ? [...contractsWithSorCode].reverse() : null),
```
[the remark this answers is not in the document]
AlexpdHackney: This has since been changed slightly, but it is so that if anyone in the future wants to use contractsWithSorCode, they can do so without it automatically being reversed

src/components/BackOffice/ContractsDashboard/Contractor/ContractorView.tsx
```diff
@@ -57,7 +57,7 @@ const ContractorView = ({
   )
 
   const descendingDateContractsWithSorCode = useMemo(
-    () => (contractsWithSorCode ? [...contractsWithSorCode].reverse() : null),
+    () => contractsWithSorCode && [...contractsWithSorCode].reverse(),
```
Duslerke, line 60: image

This will still return `null` when `contractsWithSorCode` is null. If null is unwanted, you'll probably want something like:

`contractsWithSorCode ? [...contractsWithSorCode].reverse() : []`

src/components/BackOffice/ContractsDashboard/Contractor/ContractorView.tsx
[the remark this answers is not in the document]
Duslerke: For the future reference: This is the kind of case that is not worth future proofing for.

It's because if anyone wanted to use the original sorting, they would just make that tweak themselves pretty easily. While the cost of future-proofing now is added complexity.

Though, I won't ask you to change this back now as it's a nit pick really.

PR #1160, opened 2025-08-01
Try recreating the file to prevent corruption

adamtry:
Hypothesis is that after a failed image compression, the file object becomes corrupted which is why it is failing to upload.

This should mitigate that by creating a new file using the original file data instead of using the potentially corrupted file object.


adamtry:
> Interesting. What made you consider this?

I looked into reasons for file compression failure and it mentioned that failed compression can corrupt the original file.

It seems like the only thing that makes sense for why after the compression fails the upload is also guaranteed to fail, which is what the data suggests.

There's also that older hardware and software (old phones running old Android and old Chrome) will be more likely to fail to compress the files

PR #1165, opened 2025-08-05
Add compressor js with feature toggle

adamtry:
A large number of users are getting errors with file upload. Every time there's an error with file upload there's first an error with the image compression, e.g:

```
failed to compress file - using original {"name":"1000001663.jpg","size":2407627,"type":"image/jpeg"} with error {
currentTarget: [object FileReader],
isTrusted: true,
target: [object FileReader],
type: error
}
```

It's not clear why this is happening - a theory is that users are running an excessively old Android system which somehow causes the file reading to fail in the library.

As an experimental way to try to fix this, this PR adds a feature toggle to use the CompressorJS library instead of browser-image-compression. CompressorJS is older and less sophisticated, which could mean it doesn't have the same issues on the older hardware.

However, CompressorJS is slower on modern hardware because it doesn't use a Web Worker to perform the tasks in a new process. We'll have to monitor this to ensure it helps with resolving this issue and does not introduce a performance issue.

### DIFF
--- a/serverless.yml
+++ b/serverless.yml
@@ -77,6 +77,7 @@ functions:
       SENTRY_RELEASE: ${env:CIRCLE_SHA1}
       NEW_APPOINTMENT_ENDPOINT_ENABLED: ${ssm:/repairs-hub/${self:provider.stage}/enable-new-appointments-endpoint}
       FOLLOW_ON_IS_EMERGENCY_FIELD_ENABLED: ${self:custom.follow-on-is-emergency-field-enabled.${self:provider.stage}}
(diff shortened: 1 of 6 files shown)


serverless.yml
```diff
@@ -164,4 +165,8 @@ custom:
   follow-on-is-emergency-field-enabled:
     development: true
     staging: true
-    production: true
\ No newline at end of file
+    production: true
+  use-compressor-js:
+    development: true
+    staging: false
+    production: false
```
adamtry, line 172: I'll switch this to true manually after testing on dev

Duslerke (approved):

PR #1167, opened 2025-08-05
Make completeUplaod fault tolerant

adamtry:
There are some cases where users successfully compress the files and upload them, but the final completeUpload step fails, presumably due to unstable internet connection.

By retrying this lightweight operation up to 5 times this should help to mitigate the issue.


Duslerke (approved):
I had a feeling it was due to internet connection. :thinking: 
I remember the considering of poor internet reception being requirement on another project I was on.

Duslerke:
Thing is, I guess this code will only run so long as user is on the page.
Is there anything on the FE to tell user not to navigate away after 1st failure?

PR #1183, opened 2025-08-11
Fail fast if a file to upload is unreadable

adamtry:
Currently if a file is unreadable, it will still try to upload it to S3, which will fail after about a minute.

It's better to check it is readable first and fail immediately. The error will also tell the user which file is not readable.

Reasons for the file not being readable at the time of upload could include changes to the file metadata in the background such as due to Google services like Google Photos.

image

### DIFF
--- a/src/components/WorkOrder/Photos/__snapshots__/ControlledFileInput.test.js.snap
+++ b/src/components/WorkOrder/Photos/__snapshots__/ControlledFileInput.test.js.snap
@@ -62,6 +62,11 @@ exports[`ControlledFileInput component renders shows file previews 1`] = `
             src="mockResponseUrl"
           />
         </div>
+        <p
+          class="photoUploadPreview-filename"
+        >
+          test.png
+        </p>
         <button
           class="photoUploadPreview-removeButton"
           type="button"
@@ -123,6 +128,11 @@ exports[`ControlledFileInput component renders shows validation error 1`] = `
             src="mockResponseUrl"
           />
         </div>
+        <p
+          class="photoUploadPreview-filename"
+        >
+          test.png
+        </p>
         <button
           class="photoUploadPreview-removeButton"
           type="button"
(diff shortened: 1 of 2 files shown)


src/components/WorkOrder/Photos/hooks/uploadFiles/uploadFilesToS3.ts
```diff
@@ -7,12 +7,15 @@ const uploadFilesToS3 = async (
...
   try {
+    // Upload in parallel for faster performance on good internet connections
     await Promise.all(
       files.map(async (compressedFile, i) => {
-        await uploadFileToS3(compressedFile, links[i])
-        fileUploadCompleteCallback()
+        try {
+          await uploadFileToS3(compressedFile, links[i])
+        } finally {
+          fileUploadCompleteCallback()
```
adamtry, line 17: There was a bit of a bug here where sometimes it would stay on the loading screen when files failed to upload

src/components/WorkOrder/Photos/hooks/uploadFiles/index.ts
```diff
@@ -25,6 +25,20 @@ const uploadFiles = async (
...
+        try {
+          // Attempt to read the first kilobyte of the file to ensure it is valid and accessible
+          await file.slice(0, 1024).arrayBuffer()
+        } catch (err) {
+          const errorMessage = `Could not read the file "${file.name}". Please remove and re-select it. Error: ${err.message}`
+          console.error(errorMessage, err)
+          throw Error(errorMessage)
+        }
+      })
+    )
```
adamtry, line 40: This is extremely quick / almost instant

Duslerke (approved):
I still don't think I understand why it's not readable.
Like, you make a photo with your camera app, and then I guess you select the photo file path using file explorer. Like, do they delete the photos after selecting them or smth?

PR #1202, opened 2025-08-28
Show compressed image previews instead of full size

adamtry:
There can be significant slowdown on the device when selecting multiple large images currently because the full files are loaded and displayed, even though the canvas size for those files is tiny.

This performance loss may be annoying to users and be contributing towards file compression and upload errors.

### DIFF
--- a/src/components/WorkOrder/Photos/hooks/uploadFiles/cacheFile.ts
+++ b/src/components/WorkOrder/Photos/hooks/uploadFiles/cacheFile.ts
@@ -86,7 +86,6 @@ export async function setCachedFile(file: File): Promise<void> {
       return
     }
     await db.put(STORE_NAME, file, cacheKey)
-    console.log('Cached file in IndexedDB:', fileDetails(file))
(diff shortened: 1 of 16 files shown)


src/components/WorkOrder/Photos/hooks/uploadFiles/cacheFile.ts
```diff
@@ -57,9 +57,12 @@ export async function getCachedFile(originalFile: File): Promise<File | null> {
...
     if (!cachedFile) return null
 
-    console.log('Retrieved cached file:', cacheKey, fileDetails(cachedFile))
+    const originalName = cacheKey.replace('cached-', '')
+    const reconstructedFile = new File([cachedFile], originalName, {
+      type: cachedFile.type,
+    })
 
-    return cachedFile
+    return reconstructedFile
```
adamtry, line 65: Need this to preserve the name after they've been compressed

Duslerke (approved):

PR #1207, opened 2025-09-04
Fix file validation bug?

adamtry:


src/components/WorkOrder/Photos/ControlledFileInput.tsx
```diff
@@ -41,26 +42,21 @@ const ControlledFileInput = (props: Props) => {
   } = props
 
   const inputRef = useRef<HTMLInputElement | null>(null)
-  const [isCompressing, setIsCompressing] = useState(false)
-  const [totalFilesToCompress, setTotalFilesToCompress] = useState(0)
-  const [processedCount, setProcessedCount] = useState(0)
   const [previewFiles, setPreviewFiles] = useState<File[]>([])
```
adamtry, line 45: There was too much interdependent state which was probably the cause of the UI bugs - by using the actual previewFiles count as the source of truth for the progress this ensures the UI is in sync with the real progress - easier code logic too

src/components/WorkOrders/MobileWorkingCloseWorkOrderForm.tsx
```diff
@@ -104,6 +104,7 @@ const MobileWorkingCloseWorkOrderForm = ({
   // Show validation error immediately for quick feedback
   useEffect(() => {
     trigger('workOrderFileUpload')
+    trigger('followOnFileUpload')
```
adamtry, line 107: This may have been the cause of the bug on the follow-on files - it wouldn't always re-validate when files were removed after being over the limit

LBHCallumM (approved):

PR #1214, opened 2025-09-17
Cleanup cypress tests

LBHCallumM:
Remove unused lighthouse references


Duslerke (approved):
> "Remove unused lighthouse references"

Why are they unused?
Seems like they were used at some point.

PR #1221, opened 2025-09-19
Fix deployment after dependency upgrades

LBHCallumM:
I dont actually know what the root cause was. Some package that I updated must have changed how something worked. Maybe adding .yarnlock to package json. 

I also noticed that `/swc-linux-x64-musl` was added to `@next`. So that could have added to bundle size.


serverless.yml
```diff
@@ -17,32 +17,43 @@ provider:
...
+        - '!node_modules/browserslist/**'
+        - '!node_modules/caniuse-lite/**'
+        - '!node_modules/@aws-sdk/**'
+        - '!node_modules/@babel/***'
+        - '!node_modules/sass/**'
+        - '!node_modules/@rollup/**'
+        - '!node_modules/rollup/**'
+        - '!node_modules/leaflet/**'
+        - '!node_modules/sass/**'
+        - '!node_modules/dotenv/**'
```
adamtry, line 55: Could / should these go under `exclude:` ?

LBHCallumM: idk maybe. but it works now, so im not touching it

PR #1292, opened 2026-07-22
Remove repairs finder code

LBHCallumM:
### Description of change

Doesnt look like RepairsFinder will be used any time soon. This PR can be used to add it back if necessary.


Duslerke (approved):
That's a lot of work wasted.
Why isn't it used?

